PR #13, opened 2021-02-17
Adding travis.yml file For build integration 

sherrif10:
Kindly review  thanks
 cc @k-joseph  @dkayiwa


sherrif10:
This will fail because the module doesn't run successful due to some constraint that is being worked up on, so ideally its good to merge it and continue resolving those failures cc @k-joseph  @dkayiwa

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: java
+
```
kaweesi, line 2: We are at the moment integrating directly with bamboo ci and haven't included travis on our architecture

sherrif10: ohh thanks for your feedback

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: java
```
kaweesi, line 1: add the commands to run the tests

sherrif10: @kaweesi  Have addressed and updated the pull request

sherrif10:
Hello @kaweesi  any  more suggestion on this

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+jdk:
+ - openjdk8
+ 
+ cache:
+  directories:
+  - $HOME/.m2
+
+script:
+  - mvn clean install
+  - mvn clean test
```
kaweesi, line 12: these 2 are running tests and so no need of the second

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+
+notifications:
+  email: false
+  irc:
+    channels:
+      - "irc.freenode.org#openmrs"
```
kaweesi, line 18: we don't need to notify IRC at every PR

.travis.yml
kaweesi: can you lauch a docker ref app instance and point the tests properties to it in here

kaweesi (changes requested):
thanks @sherrif10 for this, please checkout my reviews in comments, thanks

sherrif10:
Thanks let me resolve them

.travis.yml
sherrif10: Resolved thanks

.travis.yml
sherrif10: Some more clarification about this, Are you meaning of testing this commit against running docker  refapp instance

sherrif10: This may not work when trying to test it out , not sure if i answer your question thanks

kaweesi: you need the first of those 2 commands. and yes, we need to lauch ref app within travis, wait fo rit and test against it

sherrif10: Have only included only one command thanks

sherrif10: Seems am mistakenly understanding your point, isnt this only testing the build only

sherrif10:
cc @kaweesi  any more suggestions on this commits

kaweesi:
BTW, could you first fix distro's master builds, currently the tests are skipped, start from fixing that so that the ui tests run in travis without issues in distro repo

sherrif10:
> currently the tests are skipped

sure though i thought this ticket was to handle any build that is pushed via  any commit  in qa-contrib, 


> start from fixing that so that the ui tests run in travis without issues in distro repo

 sure thanks am handling this in on of the ticket am currently working on

PR #19, opened 2021-03-22
RATEST-109: Supporting running qaframework tests on github Actions

kaweesi:


.github/workflows/qa.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Set up Maven
+        uses: aahmed-se/setup-maven@v3
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
+        with:
+          maven-version: 3.6.1
```
ibacher, line 26: Do we explicitly need to set the Maven version? Otherwise, the GitHub actions images already have Maven installed.

.github/workflows/qa.yml
```diff
@@ -0,0 +1,47 @@
+name: Run QA Framework tests on Github Actions
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
ibacher, line 6: I recommend having a `workflow_dispatch:` action as well, as this allows you to rebuild via a UI click.

.github/workflows/qa.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: mvn clean install
+      - name: Checkout qaframework
+        uses: actions/checkout@master
+        with:
+          repository: openmrs/openmrs-contrib-qaframework
+      - name: Run qaframework on selenium
+        run: |
+          echo "cucumber.publish.enabled=true" > src/test/resources/cucumber.properties
+          cp -f src/test/resources/org/openmrs/uitestframework/test-travis-ci.properties src/test/resources/org/openmrs/uitestframework/test.properties
+          mvn clean install
```
ibacher, line 47:
```suggestion
          mvn -B clean install
```

`-B` enables batch-mode, which is useful when building on CI (it doesn't stop on errors, among other things). Note, too that Travis's default Maven build is actually:

```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn test -B
```

If you want to replicate the flow exactly (hence why I had [two steps](https://gist.github.com/ibacher/d8ad221092be465dcef9a4ec6434c012#file-travis-gh-sh-L44) in my example).

.github/workflows/qa.yml
kaweesi: that's awesome, i didn't imagine it's installed by default and didn't want to end up with maven 2

.github/workflows/qa.yml
kaweesi: i like option -B and will have it in, does github actions cache and parhaps i need to manage the cache in the config too!

ibacher: Caching the Maven repo needs a different action, but I've used [this one](https://gist.github.com/ibacher/56d7535b398b9716b886a27eaed7e62d#file-maven-yml-L29) before with some success.

PR #26, opened 2021-05-25
OCLOMRS-995: Automated tests for login and logout OCL

hadijahkyampeire:
[OCLOMRS-955](https://issues.openmrs.org/browse/OCLOMRS-995)

* Cucumber tests for Login and Logout


hadijahkyampeire:
@k-joseph @ibacher here is the PR, please review and unblock me.

src/test/resources/org/openmrs/uitestframework/test-ocl.properties
```diff
@@ -0,0 +1,5 @@
+webapp.url=http://localhost:8080
+login.username=haddy315
+login.password=mamanamutebi315
```
sherrif10, line 3: Thanks @hadijahkyampeire  for working on this, i think we need to make this  more generic to specific crendentials , Suppose another tester come up  with different username or password, we would not have something general, can we use admin, Admin123 for password and username

hadijahkyampeire: You are right @sherrif10 , I did that when we had just changed the environment to staging and didn't have uniform credentials but now we have so that will change.

ibacher:
@hadijahkyampeire We got the ok from both @gracepotma and @k-joseph to move these into the OCL for OpenMRS client itself, so let's see if this works better over there. In general, I'd like to avoid embedding any credentials whatsoever into test suites. Those credentials should (ideally) be stored as run-time provided properties in our CI environments (i.e., GitHub Secrets and the equivalent in Bamboo).

kdaud:
Based on the discussion [here](https://github.com/openmrs/openmrs-ocl-client/pull/702#issuecomment-857684172), there might be a need to close this PR. 
cc: @hadijahkyampeire

ibacher:
Closing this as we're keeping the ODM tests with the ODM code.

PR #39, opened 2021-06-17
MF-532: Implement an E2E test for the login flow of refapp-3.x

jayasanka-sack:
## Purpose
The purpose of this pr is to fix [MF-532](https://issues.openmrs.org/browse/MF-532)

## Goals
To give users confidence that the Login workflow of the RefApp 3.x is working as expected.

## Approach
- Create a new property file to store properties
- Remove step definition files of the ocl client (Because it has been already moved to a separate project)
- Implement an E2E test for the login flow of refapp-3.x
- Create a GitHub workflow file
- Increase the max waiting time to 40 seconds (Sometimes the test server takes that much time to load the inner components but it pretends the page has been fully loaded. Therefore increasing the waiting time is essential)
- Add workflow badge to the readme file

## Demo
Kapture 2021-07-12 at 22 59 28

### DIFF
--- a/cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
+++ b/cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
@@ -1,0 +1,35 @@
+import {Given} from 'cypress-cucumber-preprocessor/steps';
+
+Given('user arrives at the login page', () => {
+    cy.visitPage('/login');
+})
+
+When('the user logs in with {string} and {string} to the {string}', (username, password, location) => {
+    cy.getByLabel('Username').type(username)
+    cy.contains('Continue').click();
+    cy.getByLabel('Password').type(password)
+    cy.contains('Log in').click();
+    // The user will be redirected to the location selection page if the credentials are correct.
+    // The following artificial wait is to wait until it changes the url.
+    cy.wait(2000);
+    // If the url includes "/location" that means the user is on the location selection page. If so, pick the location.
+    cy.url().then(($url) => {
+        if($url.includes("/location")) {
+            cy.contains(location).click();
+            cy.contains('Confirm').click();
+        }
+    })
+})
+
+Then('the user should be {string} to login', (ability) => {
+    switch (ability) {
+        case "able":
+            cy.url().should('include', '/home');
+            break;
+        case "unable":
+            cy.contains("Incorrect username or password");
+            break;
+        default:
+            throw new Error(`Ability '${ability}' is not supported`);
+    }
+});
--- a/src/test/resources/org/openmrs/uitestframework/test-cypress.properties
+++ b/src/test/resources/org/openmrs/uitestframework/test-cypress.properties
@@ -1,0 +1,1 @@
+webapp.url=https://openmrs-spa.org/openmrs/spa
--- a/.github/workflows/refapp-3x-login.yml
+++ b/.github/workflows/refapp-3x-login.yml
@@ -1,0 +1,17 @@
+name: RefApp 3.x Login
+on:
+  pull_request:
+    branches: [ master ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Using Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Installing dependencies
+        run: npm install
+      - name: Run patient registration workflow tests
+        run: npm run refApp3Login
--- a/cypress.json
+++ b/cypress.json
@@ -4,5 +4,6 @@
     "*.feature",
     "**/*.feature"
   ],
-  "video": false
+  "video": false,
+  "defaultCommandTimeout": 40000
 }
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,6 +1,6 @@
 const cucumber = require('cypress-cucumber-preprocessor').default
 var propertiesReader = require('properties-reader');
-var testProperties = propertiesReader('src/test/resources/org/openmrs/uitestframework/test.properties');
+var testProperties = propertiesReader('src/test/resources/org/openmrs/uitestframework/test-cypress.properties');
 
 module.exports = (on, config) => {
   on('file:preprocessor', cucumber());
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -20,3 +20,17 @@ Cypress.Commands.add('initiateExceptionsLogger', () => {
     	return false;
     });
 });
+
+Cypress.Commands.add('visitPage', path => {
+    cy.task('getProperty', 'webapp.url').then(baseUrl => {
+        cy.visit(`${baseUrl}${path}`);
+    });
+})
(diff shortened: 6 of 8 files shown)


src/test/resources/org/openmrs/uitestframework/test-cypress.properties
```diff
@@ -0,0 +1 @@
+webapp.url=https://openmrs-spa.org/openmrs/spa
```
jayasanka-sack, line 1: Is this the baseurl that are we going to use? 

Will it be suitable for the test that has "create" operations? (ex: patient registration)

jayasanka-sack:
@kaweesi Could you please review this? 😇

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
```diff
@@ -0,0 +1,24 @@
+import {Given} from 'cypress-cucumber-preprocessor/steps';
+
+Given('user arrives at the login page', () => {
+    cy.task('getProperty', 'webapp.url').then(baseUrl => {
```
kaweesi, line 4: am wondering if we should extract out a re-usable function/command for visiting a url. otherwise this looks nice

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
```diff
@@ -0,0 +1,24 @@
...
+        cy.visit(`${baseUrl}/login`);
+    });
+})
+
+When('the user logs in with {string} and {string}', (username, password) => {
+    cy.get('#username').type(username);
+    cy.contains('Continue').click();
+    cy.get('#password').type(password);
+    cy.contains('Log in').click();
+    cy.wait(3000);
```
kaweesi, line 14: any reason why you want to delay here?

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
kaweesi: yeah, that should be the base-url

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
jayasanka-sack: Yeah that would be nice! I'll do it!

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
jayasanka-sack: Actually, it takes a long time to load. But later on, I figured out a way to increase the waiting time (by updating the cypress json file). I'll update the PR

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
jayasanka-sack: Fixed.

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
jayasanka-sack: fixed.

jayasanka-sack:
@kaweesi, @brandones  I changed the structure of the workflow (inspired by a previous implementation on ref app 2.x) 
Also added a GitHub workflow as well. Could you please review this when you are not busy?

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
```diff
@@ -0,0 +1,35 @@
...
+})
+
+When('the user logs in with {string} and {string} to the {string}', (username, password, location) => {
+    cy.get('#username').type(username);
+    cy.contains('Continue').click();
+    cy.get('#password').type(password);
+    cy.contains('Log in').click();
+    // The user will be redirected to the location selection page if the credentials are correct.
+    // The following artificial wait is to wait until it changes the url.
+    cy.wait(2000);
```
brandones, line 14: Don't use static `wait`s. Please see [the Cypress documentation](https://docs.cypress.io/faq/questions/using-cypress-faq#How-can-I-make-Cypress-wait-until-something-is-visible-in-the-DOM).

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
```diff
@@ -0,0 +1,35 @@
+import {Given} from 'cypress-cucumber-preprocessor/steps';
+
+Given('user arrives at the login page', () => {
+    cy.visitPage('/login');
+})
+
+When('the user logs in with {string} and {string} to the {string}', (username, password, location) => {
+    cy.get('#username').type(username);
```
brandones, line 8: Don't get elements by ID. IDs are implementation details. Get elements by how they appear to the user. In this case, it is a text input with label `Username`.

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
brandones: I thought that we were going to be running tests on isolated test servers.

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
kaweesi: do you have a plan to maintain the ids permanent? if so then i would strongly think ids/classes etc are a great ways to refer to elements vs labels which may change with locale changes etc which could simply be out of the test scope

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
kaweesi: a test instance for microfrontends should be started in the github actions and waited for before running tests whose url should used here insyead

brandones: Great. I think that @jayasanka-sack will need some help making that happen, and that we probably shouldn't block this PR on it.

cypress/integration/cucumber/step_definitions/refapp-3.x/01-login/login.js
brandones: No; ids and classes are implementation details. "Implementation details" are aspects of the code which are invisible to the user. [Do not test implementation details](https://kentcdodds.com/blog/testing-implementation-details#why-is-testing-implementation-details-bad). We should test against the English labels. Those are the "source" labels in the code. If those change, it will be highly apparent, and easy to adapt the tests. AFAIK locale changes are not currently in test scope; but testing them is not out of the question.

brandones: This is why I have recommended we use [Cypress Testing Library](https://testing-library.com/docs/cypress-testing-library/intro/)—it provides very natural semantics for "testing from the user's perspective." See also the [Testing Library Introduction](https://testing-library.com/docs/). This video, "[Build invincible tests](https://www.youtube.com/watch?v=F8LH9d9eN3M)," is also very good.

brandones (approved):

PR #40, opened 2021-06-19
RATEST-T: Workflow based testing framework

kdaud:


sherrif10:
Try to make commit messages in full names to make it more readable

kdaud:
> Try to make commit messages in full names to make it more readable

I agree. 

Thanks @sherrif10  for  the review. However, this commit is a `node` inheriting traits of its parent.  When its all constructed and set, I will then update the description and as well migrate its respective ticket state to `READY FOR REVIEW`  for reviewers to take action.

PR #46, opened 2021-07-07
Add Gecko driver for macOS

ibacher:


pom.xml
```diff
@@ -85,7 +85,19 @@
                         <phase>process-test-classes</phase>
                         <configuration>
                             <target>
-                                <chmod file="target/test-classes/firefoxdriver/linux/geckodriver" perm="755" />
+                                <chmod file="target/test-classes/firefoxdriver/mac/geckodriver" perm="755"/>
```
kaweesi, line 88: i think we can add failOnError set to false to allow other OS to run in-case the file is missing, like in CI. 

https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/pom.xml#L107

ibacher: @kaweesi Good call. I've actually just added that for every driver.

PR #50, opened 2021-07-25
RATEST-169: Amend RefApp 3.x Login E2E to run in a dockerized environment

jayasanka-sack:
## Purpose
The purpose of this PR is to fix  https://issues.openmrs.org/browse/RATEST-169

## Goals
- Make the tests repeatable despite data mutations

## Approach
- Reuse `docker/docker-compose-refqaa.yml` composer file to fire up a docker container
- Run 3.x frontend on top of the instance
- Change the base url


.github/workflows/refapp-3x-login.yml
```diff
@@ -1,17 +1,43 @@
...
+      - name: wait for openmrs instance to start
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 1; done
       - name: Using Node.js
         uses: actions/setup-node@v1
         with:
           node-version: '12.x'
       - name: Installing dependencies
         run: npm install
-      - name: Run patient registration workflow tests
+      - name: Run RedApp 3.x frontend
```
ibacher, line 39:
```suggestion
      - name: Run RefApp 3.x frontend
```

.github/workflows/refapp-3x-login.yml
```diff
@@ -1,17 +1,43 @@
...
+        run: npx openmrs@next start --backend localhost:8080 &
```
ibacher, line 40: So here's where things might get a little bit difficult. If the backend is already running on port 8080, then we can't have the frontend running on port 8080. Only one executable can claim one port at a time. So, either we need a new docker-compose file that runs the backend on a port other than 8080 (e.g., 8081) or we need to run the frontend on a port other than 8080 and ensure that Cypress connects to that port.

ibacher (changes requested):

.github/workflows/refapp-3x-login.yml
ibacher: PS The way we make things work in the 3.x distro is precisely this. We have three web Docker components:

* The 3.x UI running on one server
* The OpenMRS backend running on another server
* An entrance proxy running on a third server, which keeps everything tied to the same port.

You can see the details of how this works by examining the setup [here](https://github.com/openmrs/openmrs-distro-referenceapplication/tree/3.x/run/docker). The key file for making everything appear to be running on the same port is [here](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/3.x/run/docker/proxy/proxy.conf). Note how the frontend and backend both run on 8080, but on different "hostnames".

jayasanka-sack: Oh! Thanks, Ian. This is very helpful!

I tried running the proxy on a different port. The UI was there but it didn't load any data.
the following error occurs, but let me update the workflow and try it.
```
[HPM] Error occurred while proxying request localhost:8081/openmrs/ws/rest/v1/session to localhost:8080 [EHOSTUNREACH] (https://nodejs.org/api/errors.html#errors_common_system_errors)
[HPM] Error occurred while proxying request localhost:8081/openmrs/ws/rest/v1/session to localhost:8080 [EHOSTUNREACH] (https://nodejs.org/api/errors.html#errors_common_system_errors)
[HPM] Error occurred while proxying request localhost:8081/openmrs/ws/rest/v1/visit?includeInactive=false&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime) to localhost:8080 [EHOSTUNREACH] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

jayasanka-sack: @ian @brandones,
I changed the frontend to run in port 8081 (`--port 8081`) also changed the port in cypress.
Still no luck.

.github/workflows/refapp-3x-login.yml
```diff
@@ -1,17 +1,43 @@
...
+        image: selenium/node-firefox:3.141.59-gold
+        env:
+          HUB_HOST: hub
+          HUB_PORT: 4444
     steps:
-      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
```
ibacher, line 23: Do we actually need to load Java here? (The other tests are actually run through Maven, but these tests are all Cypress based, right?)

.github/workflows/refapp-3x-login.yml
```diff
@@ -1,17 +1,43 @@
...
+  workflow_dispatch:
 jobs:
   build:
     runs-on: ubuntu-latest
+    services:
+      firefox:
+        image: selenium/node-firefox:3.141.59-gold
+        env:
+          HUB_HOST: hub
+          HUB_PORT: 4444
```
ibacher, line 18: Do we need to install the Selenium Firefox image for this build to work?

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
```diff
@@ -1 +1 @@
-webapp.url=https://openmrs-spa.org/openmrs/spa
+webapp.url=http://localhost:8081/openmrs/spa
```
ibacher, line 1: Does this file actually get picked up anywhere? Cypress is usually configured either via the command line or the `cypress.json` file.

.github/workflows/refapp-3x-login.yml
```diff
@@ -1,17 +1,43 @@
...
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 1; done
       - name: Using Node.js
         uses: actions/setup-node@v1
         with:
           node-version: '12.x'
       - name: Installing dependencies
         run: npm install
-      - name: Run patient registration workflow tests
+      - name: Run RefApp 3.x frontend
+        run: npx openmrs@next start --backend localhost:8080 --port 8081 &
```
brandones, line 40: Per discussion on Slack,
```suggestion
        run: npx openmrs@next start --backend http://localhost:8080 --port 8081 &
```

.github/workflows/refapp-3x-login.yml
jayasanka-sack: Yeah. No need at all. removed it.

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
jayasanka-sack: Moved it to a properties file as per the instructions given by Kaweesi. The intention was to keep all the properties in a single place. The file is been loaded in `cypress/plugins/index.js` file.  An additional cypresses task is written `cypress/support/commands.js` to visit a page.

However, having the base URL configured in the  `cypress.json` seems pretty straightforward. What do you think?

.github/workflows/refapp-3x-login.yml
jayasanka-sack: Thanks @brandones ! 
I have done the necessary changes. It works fine in my local setup. But the following issue occurs in the Github workflow:
`Error: connect ECONNREFUSED 127.0.0.1:8081`


also, I tried running the test and the proxy in a single step as well. But the error is the same.
```
- name: Run patient login workflow tests
        run: |
                npx openmrs@next start --backend http://localhost:8080 --port 8081 &
                npm run refApp3Login
```

brandones: I think [this SO Q/A](https://stackoverflow.com/a/62546833/1464495) is about this same issue. It seems like we need to publish the image and use it as a service container: https://docs.github.com/en/actions/guides/about-service-containers

jayasanka-sack: Thanks @brandones,

Necessary changes were done to the workflow file:
https://github.com/openmrs/openmrs-contrib-qaframework/pull/50/commits/d522a0f602965f5c44af6a3db3f1f5a1c4dbe31b

Had to remove some parameters such as `command` `healthcheck` and `depends_on` because Github doesn't support them. 

Now it throws the following error:
```
Exit code 1 returned from process: file name '/usr/bin/docker', arguments 'create --name ce9c6e7005dd4f9bba8f5188bffc9c5d_openmrsopenmrsreferenceapplicationdistrodemo_516d6f --label 3bec5f --network github_network_4f55a29bcb3942ffa425949bc63bcea0 --network-alias openmrs-referenceapplication -p 8080:8080  -e "DB_DATABASE=openmrs" -e "DB_HOST=openmrs-referenceapplication-mysql" -e "DB_USERNAME=openmrs" -e "DB_PASSWORD=Admin123" -e "DB_CREATE_TABLES=true" -e "DB_AUTO_UPDATE=false" -e "MODULE_WEB_ADMIN=false" -e GITHUB_ACTIONS=true -e CI=true -v "openmrs-referenceapplication-data":"/usr/local/tomcat/.OpenMRS" -v "./activemq-data":"/usr/local/tomcat/.OpenMRS/activemq-data" openmrs/openmrs-reference-application-distro:demo'.
```

https://github.com/openmrs/openmrs-contrib-qaframework/runs/3184870317?check_suite_focus=true

jayasanka-sack: Btw, I'm curious how 2.x tests access the localhost. 

However, I think the container runs fine because `Wait for Openmrs instance to start` stage in the workflow works fine.

Can't it be an issue with the frontend proxy? 🤔

brandones: Looks like the actual error is `Error response from daemon: create ./activemq-data: "./activemq-data" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path`. So `./activemq-data` is not okay because it's a relative path. Try making it an absolute path.

brandones: Yeah I have the same curiosity about the 2.x tests. They definitely do work. Not sure why Selenium is able to access the container but we're not able to set up a proxy to it.

Anyway let's keep on with the `services` approach, this seems promising.

jayasanka-sack: Thanks @brandones , I have fixed the path. The container is now running. However, it throws the `Error: connect ECONNREFUSED 127.0.0.1:8081` error again. What could go wrong?

brandones: I don't know, I'm not an expert in GitHub Actions. I need you to research this problem using Google (or another search engine) and on Stack Overflow and try things until you are able to solve it. That is what I would do. I don't think that we are trying to do something impossible. But if you decide that we need to take a different approach, please lead us in that direction.

src/test/resources/org/openmrs/uitestframework/test-cypress.properties
brandones: Yes, I think configuring it in `cypress.json` is fine.

brandones:
Okay, some thoughts:

- As discussed, first thing to do is add a curl to find out if the frontend server is working and whether the proxy is working, prior to running the tests. For the frontend server, `curl` a page. For the proxy, `curl` a GET API endpoint that the application uses.
- You actually should see if you can reproduce this locally by running the OpenMRS server and MySQL in docker compose and then running the frontend and the tests from the command line. This would be very informative. It would tell us whether or not the issue has to do with GitHub Actions.
- If it turns out that the frontend server is at fault...
    - Try using the [GitHub Action for npx](https://github.com/marketplace/actions/github-action-for-npx)?
    - This would be both surprising and bad.
- If it turns out that the proxy is at fault...
    - You could try changing the port that the OpenMRS docker container exposes from 8080 to, e.g., 8081, and let the frontend run at its default port.
    - You could try changing `localhost` to `127.0.0.1` or another synonym, for `--backend`
    - You should know that "http proxy middleware" is the NPM package that makes the proxy happen
- If it turns out that the Cypress connection is at fault...
    - Try fiddling with whatever Cypress connection parameters are available
    - Try using the [Cypress GitHub Action](https://github.com/cypress-io/github-action) (see also [here](https://docs.cypress.io/guides/continuous-integration/github-actions#Cypress-GitHub-Action))

jayasanka-sack:
@brandones I just added two curl commads to test both openmrs instance and the proxy.
The openmrs instance works fine, but the proxy doesn't. 
https://github.com/openmrs/openmrs-contrib-qaframework/runs/3204190475?check_suite_focus=true

Thanks for your suggestions. It's a bit late here. I'll try them tomorrow morning and let you know. 
Thanks again!

brandones:
This is testing the backend server and the frontend server:

```
          curl http://localhost:8080  # backend server
          sleep 30
          curl http://localhost:8081  # frontend server
```

Let's try

```
curl http://localhost:8081/openmrs/spa/home  # test the frontend
curl http://localhost:8081/openmrs/ws/rest/v1/session  # test the proxy
```

It probably doesn't matter, but run each command in its own step. Easier to understand the output and move things around that way.

And then yeah, follow my suggestions above for ways to continue investigating this, now that we understand the distinction between the frontend server and the proxy.

jayasanka-sack:
@brandones 

These are the experiments I've done:
* Run as services
		* curl http://localhost:8080
			* 200 OK
		* curl http://localhost:8081/openmrs/spa/home
			curl: (7) Failed to connect to localhost port 8081: Connection refused
	
		* curl http://localhost:8081/openmrs/ws/rest/v1/session
			curl: (7) Failed to connect to localhost port 8081: Connection refused
* switch the ports
		* instance is running:
		* curl: (7) Failed to connect to localhost port 8081: Connection refused
* Ran the docker container locally.
		* Was able to get it up and running after a few attempts. May be because I'm running it on a M1 macbook. (But it fails to start again later)
		* The frontend doesn't get loaded. It shows the loading animation then it shows the following message.
			* "Oops! An unexpected error occurred."
		* Uncaught Stacktrace for outgoing request:
		* Error: Server responded with 404 (Not Found) for url /openmrs/ws/rest/v1/session. Check err.responseBody or network tab in dev tools for more info
			* Seems the web services module doesn't existn on the installation
			* curl http://localhost:8081/openmrs/ws/rest/v1/session
			* 404 not found
		* Seems web services module isn't there.
* Testing https://github.com/openmrs/openmrs-distro-referenceapplication/tree/3.x locally
		* Couldn't start the instance
		* org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'conversionService' defined in URL [jar:file:/openmrs/data/.openmrs-lib-cache/uiframework/uiframework.jar!/webModuleApplicationContext.xml]: Initialization of bean failed; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'stringToAppDescriptorConverter': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanNotOfRequiredTypeException: Bean named 'appFrameworkService' is expected to be of type 'org.openmrs.module.appframework.service.AppFrameworkService' but was actually of type 'com.sun.proxy.$Proxy326'
* Testing https://github.com/openmrs/openmrs-distro-referenceapplication/tree/3.x in a Github workflow
		* https://github.com/openmrs/openmrs-contrib-qaframework/runs/3213518157?check_suite_focus=true
		* OpenMRS instance is working
		* Running curl http://localhost/openmrs/spa/home throws a 500
		<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN"> 
		13<html><head> 
		14<title>500 Internal Server Error</title> 
		15</head><body> 
		16100   531  100   531    0     0   172k      0 --:--:-- --:--:-- --:--:--  172k 
		17<h1>Internal Server Error</h1> 
		18<p>The server encountered an internal error or 
		19misconfiguration and was unable to complete 
		20your request.</p> 
		21<p>Please contact the server administrator at 
		22 [no address given] to inform them of the time this error occurred, 
		23 and the actions you performed just before this error.</p> 
		24<p>More information about this error may be available 
		25in the server error log.</p> 
		26</body></html>
* Testing https://github.com/openmrs/openmrs-distro-referenceapplication/tree/3.x-npx-no-install in a Github workflow
		* Branch 3.x-npx-no-install
		* fails at OpenMRS package
			* Failed to execute goal on project openmrs-distro-package: Could not resolve dependencies for project org.openmrs:openmrs-distro-package:pom:1.0.0-SNAPSHOT: Failed to collect dependencies at org.openmrs.web:openmrs-webapp:war:2.4.0: Failed to read artifact descriptor for org.openmrs.web:openmrs-webapp:war:2.4.0: Could not transfer artifact org.openmrs.web:openmrs-webapp:pom:2.4.0 from/to maven-default-http-blocker (http://0.0.0.0/): Blocked mirror for repositories: [openmrs-repo (http://mavenrepo.openmrs.org/nexus/content/repositories/public, default, releases+snapshots)] -> [Help 1]

brandones:
Okay, thanks @jayasanka-sack for such a clear and detailed report.

For the server issue—I'm pretty sure that the Refapp 2.x Docker image contains the REST WS module. See https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/package/src/main/resources/openmrs-distro.properties . @ibacher , have I got that right?

In which case @jayasanka-sack I'm not sure why `/openmrs/ws/rest/v1/session` would 404. First try it directly against the server, then against the proxy. It should work for both.

In your latest run it looks like you're trying to mitigate the connection issues with the frontend server by instead having the 3.x server serve the frontend. This is cool and maybe could work, but the 3.x server is still AFAIK not in working shape, and also it would probably require building a new Docker image for each run, which is a lot more overhead than just running the frontend dev server.

Please keep working on getting the backend in place, with the REST WS and all. I'll look into the frontend connection issues more.

ibacher:
> have I got that right?

Yes, although we have had reports that REST 2.31.0 sometimes has issues running on platform 2.4.0. Updating REST usually fixes that, though that isn't easy with a Docker image...

brandones:
How not easy? Any chance you could outline the process such that @jayasanka-sack would be able to take that on, if he's able to confirm that that's indeed the problem?

ibacher:
So, actually, my best guess right now is that with the RefApp 2.x image, this sequence of commands isn't waiting long enough for the RefApp to actually be started (it has to go through the original page load). It might actually be worthwhile to use the return value of the requests to `http://localhost:8080/openmrs/ws/rest/v1/session` to check whether OpenMRS is started rather than looking for the login page (the problem is that if OpenMRS is starting, it will just serve you the initialsetup page, but it looks like a sucessful HTTP request to CURL).

jayasanka-sack:
Thanks for your inputs, @ibacher and @brandones,

I configured the workflow to print the status code of  `http://localhost:8080/openmrs/ws/rest/v1/session` 
it prints 302 until the instance starts (like 3 minutes) and then start printing 404. I think the issue with the module. 

https://github.com/openmrs/openmrs-contrib-qaframework/runs/3231652058?check_suite_focus=true

brandones:
Okay, I'm working on debugging this. I'll post an update before I end my day today. Based on that, feel free to keep debugging during my-night-your-day, or work on debugging why connections to the frontend server fail.

sherrif10:
Hello @jayasanka-sack.   Is this work still valid, Can we try to resolve the merge conflicts.

PR #54, opened 2021-07-28
RATEST-150: E2E Search and Registration of New Patient

sherrif10:
https://issues.openmrs.org/browse/RATEST-150 cc @ibacher @dkayiwa @kdaud


sherrif10:
cc @dkayiwa , still we have same issue here https://github.com/openmrs/openmrs-contrib-qaframework/pull/54/checks?check_run_id=3192178653

kdaud:
> still we have same issue here

@sherrif10 as I mentioned [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/516#issuecomment-888801773) on this [PR](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/516) did you test locally this workflow before the sister was merged?

dkayiwa:
> in the sister PR

and the brother pr?

sherrif10:
This commit work perfect , All the automations are working as expected locally , the only problem is that github actions and configured ci tracks all the changes being pushed,this commit only breaks after pushing  with the same issue error log.

dkayiwa:
@sherrif10 the error is different from the original one.

kdaud:
> the only problem is that github actions and configured ci tracks all the changes being pushed,this commit only breaks after pushing with the same issue error log.

When the test passes well locally then it should pass on Actions as well. You might need to consider testing locally the workflow on firefox and chrome for both headless true/false instance

sherrif10:
> > the only problem is that github actions and configured ci tracks all the changes being pushed,this commit only breaks after pushing with the same issue error log.
> 
> When the test passes well locally then it should pass on Actions as well. You might need to consider testing locally the workflow on firefox and chrome for both headless true/false instance

So you mean when i say it passes locally i meant it fails locally

sherrif10:
Waiting on the build thanks for reviewing ,its my honor

kdaud:
> So you mean when i say it passes locally i meant it fails locally

@sherrif10 what I meant is to ensure the test workflow is tested on chrome headless false and true, and as well on firefox instance too!

Have had use cases where the test could pass locally  on firefox headless `true` and `false` and as well passes on chrome headless `false` and **fails** with chrome headless `true`. Not until have debugged the test to pass on all instances then ci passes too! Otherwise that is what I meant in particuaar

sherrif10:
> @sherrif10 the error is different from the original one.

Hello @dkayiwa  it still the same error as seen through the logs here   Searching for missing patient(Search And Patient Registration): Expected condition failed: waiting for element to be clickable: By.id: phoneNumberLabel ()

kdaud:
> Expected condition failed: waiting for element to be clickable: By.id: phoneNumberLabel ()

Does this mean when you use `visibility` rather than `clickable` the test pass?

sherrif10:
> s mean when you use `visibilityOfElement` rather than `clickable` the test passes?

Thats what am thinking currently to be the last option since am seeing logs are the returning the same error

kdaud:
> Thats what am thinking currently to be the last option since am seeing logs are the returning the same error

As @dkayiwa mentioned [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/516#discussion_r678966369) I think the solution  is neither using `visibilityOfElement` nor `clickable ` but rather looking into why the element is not visible. 

Some pages takes longer to load and the mess might be caused by the loading of the previous page!

sherrif10:
Carrying out a quick test on bamboo, Watching all the builds and failures of this project

sherrif10:
Reverted this

PR #65, opened 2021-08-09
RATEST-182: Adding chromedriver and geckodriver for windows os

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-182)
Description -> Add chromedriver and geckodriver for windows os


kdaud:
Am skeptical whether windows may need more configurations for the driver to work. Ideally one will just need to download and replace the firefox or chrome driver version which is compatible with the browser version running on the machine

sherrif10:
> Am skeptical whether windows may need more configurations for the driver to work. But ideally one will just need to download and replace the firefox or chrome driver version which is compatible with the browser version running on the machine and as well changing the `webdriver.gecko/chrome.driver` location value in the `test.properties` file to point to the right location

According to your statements above, Is this ticket still valid ?

kdaud:
> > Am skeptical whether windows may need more configurations for the driver to work. But ideally one will just need to download and replace the firefox or chrome driver version which is compatible with the browser version running on the machine and as well changing the `webdriver.gecko/chrome.driver` location value in the `test.properties` file to point to the right location
> 


Such a statement ideally implies that the PR author is open to ideas and perhaps hearing experiences of any one familiar with windows os behavior in reference to selenium web drivers.  As I was waiting for feedback on the caption I raised, I took the the initiate to have a simulation test on windows os to clear my skepticism before the commit is merged. Though I did the simulation on windows 10 os and drivers were working fine without any other configurations, But not sure whether windows 7 os will behave the same way and BTW windows 11 is almost on the door if not already out and I am putting all this in mind. 

@sherrif10 do you have an experience with windows os in respect to selenium web driver behavior on other windows versions?

sherrif10:
Will try to simulate an automation on a windows machine Later however still i think thats fine with adding chrome drivers for windows though it would be so testable if someone is testing it on windows machine

kdaud:
cc: @dkayiwa

PR #68, opened 2021-08-11
create admit patient test

gracebish:
create admitPatient.yml file

### DIFF
--- a/admitPatient.yml
+++ b/admitPatient.yml
@@ -1,0 +1,52 @@
+name: RefApp 2.x admit patient Test
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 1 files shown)


HerbertYiga:
thanks @gracebish  do you have any ticket id for this?

sherrif10:
We might not neccesariy need to define the whole configuration of the hook , we only need to use the already defined @selenium hook, by doing this config, we are introducing an E2E  test based workflow and creating a new run method.

kdaud:
@gracebish provide the ticket url to this commit.

gracebish:
Thank you @sherrif10. Let me close this.

PR #69, opened 2021-08-12
RATEST-187: Refactor refapp-2x test workflows configurations to a more clear names

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-187)
Description -> Refactor refapp-2x test workflows configurations to a more clear names

### DIFF
--- a/.github/workflows/refapp-2x-clinical-visit.yml
+++ b/.github/workflows/refapp-2x-clinical-visit.yml
@@ -47,5 +47,5 @@ jobs:
           echo "cucumber.publish.enabled=true" > src/test/resources/cucumber.properties
           mvn clean install -DskipTests=true
           cp -f src/test/resources/org/openmrs/uitestframework/test-local.properties src/test/resources/org/openmrs/uitestframework/test.properties
-          npm run clinicalVisitTest
+          npm run refapp2ClinicalVisitTest
           curl -sX POST -F messages=@target/cucumber.ndjson https://studio.cucumber.io/cucumber_project/results -H "project-access-token: ${{secrets.CUCUMBER_IO_TOKEN}}" -H "provider: github" -H "repo: ${{github.repository}}" -H "branch: master" -H "revision: ${{github.sha}}"
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/Steps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/Steps.java
@@ -19,7 +19,7 @@ import static org.junit.Assert.*;
 
 // Use english locale, of not set, the test instance should be set to english language
 public class Steps extends ReferenceApplicationTestBase {
-	
+
 	protected TestProperties testProperties = TestProperties.instance();
 	protected LoginPage loginPage;
 	protected FindPatientPage findPatientPage;
--- a/.github/workflows/refapp-2x-vitals-and-triaging.yml
+++ b/.github/workflows/refapp-2x-vitals-and-triaging.yml
@@ -47,5 +47,5 @@ jobs:
           echo "cucumber.publish.enabled=true" > src/test/resources/cucumber.properties
           mvn clean install -DskipTests=true
           cp -f src/test/resources/org/openmrs/uitestframework/test-local.properties src/test/resources/org/openmrs/uitestframework/test.properties
-          npm run vitalsAndTriagingTest
+          npm run refapp2VitalsAndTriagingTest
           curl -sX POST -F messages=@target/cucumber.ndjson https://studio.cucumber.io/cucumber_project/results -H "project-access-token: ${{secrets.CUCUMBER_IO_TOKEN}}" -H "provider: github" -H "repo: ${{github.repository}}" -H "branch: master" -H "revision: ${{github.sha}}"
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "scripts": {
     "refappCypress": "./node_modules/.bin/cypress-tags run -eTAGS='@cypress and not @microfrontend'",
     "refappSelenium": "mvn test -Dcucumber.filter.tags='@selenium and not @initialSetup'",
-    "vitalsAndTriagingTest": "mvn test -Dcucumber.filter.tags='@selenium and @vitals'",
-    "clinicalVisitTest": "mvn test -Dcucumber.filter.tags='@selenium and @clinicalVisit'",
+    "refapp2VitalsAndTriagingTest": "mvn test -Dcucumber.filter.tags='@selenium and @vitals'",
+    "refapp2ClinicalVisitTest": "mvn test -Dcucumber.filter.tags='@selenium and @clinicalVisit'",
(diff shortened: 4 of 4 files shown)


src/test/java/org/openmrs/contrib/qaframework/automation/Steps.java
```diff
@@ -19,7 +19,7 @@
 
 // Use english locale, of not set, the test instance should be set to english language
 public class Steps extends ReferenceApplicationTestBase {
-	
+
```
kdaud, line 22: Removed `white spaces` to obey the principle of silence for empty lines that git admires!

kdaud:
@sherrif10 take a review and recommend or approve the PR !

src/test/java/org/openmrs/contrib/qaframework/automation/Steps.java
kdaud, line 22: Removed `white spaces` to obey the principle of silence for empty lines that git admires!

kdaud:
@sherrif10 take a review and recommend or approve !

.github/workflows/refapp-2x-clinical-visit.yml
```diff
@@ -47,5 +47,5 @@ jobs:
           echo "cucumber.publish.enabled=true" > src/test/resources/cucumber.properties
           mvn clean install -DskipTests=true
           cp -f src/test/resources/org/openmrs/uitestframework/test-local.properties src/test/resources/org/openmrs/uitestframework/test.properties
-          npm run clinicalVisitTest
+          npm run refapp2ClinicalVisitTest
```
sherrif10, line 50: Can we make it refApp_2.xClinicalVisitTest or something case sensitive.

kdaud: > Can we make it refApp_2.xClinicalVisitTest or something case sensitive. with other defined hook


I took a look at the exisiting naming convention eg for [refappSelenium](https://github.com/openmrs/openmrs-contrib-qaframework/blob/0c23df71ec10866a4c29f633444f6d8236f41500/.github/workflows/qa.yml#L50) [platformUpgrade](https://github.com/openmrs/openmrs-contrib-qaframework/blob/0c23df71ec10866a4c29f633444f6d8236f41500/.github/workflows/platform-upgrade.yml#L60) [simpleCoreInstall](https://github.com/openmrs/openmrs-contrib-qaframework/blob/0c23df71ec10866a4c29f633444f6d8236f41500/.github/workflows/platform-install-simple.yml#L59) they shouts out clearly that they actually accomplish.

I suggested we would rather have our new test workflows for refapp 2.x be renamed to such names have proposed. They shouts out clearly well like `refapp2ClinicalVisitTest` mentions clearly the` refapp` version and what the test does. I was hesitant not to put the `x` value on the integer version value since it caters for all 2.x versions so I would like to think we embrace the suggested one.

sherrif10: > refapp2ClinicalVisitTest

refApp2.xClinicalVisitTest , what do you think about that naming convention.

kdaud: > **refApp2.xClinicalVisitTest ,** what do you think about that naming convention.

I have given the reason [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/69#discussion_r687751440) why I do think we should not attach `x` value on the integer version value.

kdaud: > i think we need to have **x** on the hook to make it more clear.

Actually this is not a hook but a command that triggers a workflow, based on the reason have given [above ](https://github.com/openmrs/openmrs-contrib-qaframework/pull/69#discussion_r687751440) concerning why we should not attach the `x` value its sufficient enough not to attach `x` value on an the integer version `value`.

sherrif10:
Yeah sure it's not a hook, however it will look more clear according to
.yml file of this class, let's make it standard. Otherwise I think after
that , we need to merge this

kdaud:
> let's make it standard

I would rather propose we embrace the suggested one than embedding an `x `value on a command. The version of refapp which is `2` is good enough to distinguish a test for RefApp 2.x vs 3.x than having x embedded on a command. `It will be a bit shouting !`eg `refapp2VitalsAndTriagingTest`

kdaud:
@sherrif10  I would love to know if there are more suggestions to this PR otherwise I would like to see other geeks taking a final review!

kdaud:
I am careful to tag @dkayiwa since @sherrif10 seems occupied though I would love to tag the geeks when Sharif has approved!

dkayiwa:
@kdaud what makes you feel like he is occupied?

kdaud:
> @kdaud what makes you feel like he is occupied?

I once read a comment where @dkayiwa mentioned clearly that both of us reviewing each others work should be a priority though am not able to reference it. Though It was my custom before to review all the contributors PRs to get them progressing But again had to pay more attention to my colleagues PRs to support him progress with his technical work and its ever my priority to do that every single day whenever I can unless I am away from my computer but its my priority. 

Yes when @dkayiwa gave a guidance to both of us, surely @sherrif10 pulled up on reviewing my PRs too. However, I notice (`some times`) he takes longer to review/reply my work and keeps me a bit behind technically and would love as well to progress. 

Yes I know one may be occupied with other work elsewhere since we have an array of tasks as geeks but would like to suggest both of us prioritize in giving a quick response to each other PRs to progress together. I am care Sharif is reviewing my work But some times I wait for feedback for over 6 hours or 4 hours or 3 hours or 2 hours which I see should be improved. Yes @sherrif10 is doing good but can as well do it better.

cc: @dkayiwa @sherrif10

dkayiwa:
What i mean't with priority was not necessarily meaning leaving whatever one is doing to first review a new pull request. Of course it would be great whenever possible. But sometimes it is not. So in such cases where it is not possible, taking 2hrs, or 3hrs, or even at times 6hrs, sounds reasonable and is still within my expectation of priority.

kdaud:
> Are you putting these tests in the same place such that you need to differentiate them by name?

@dkayiwa after taking time to look through the existing workflows within the module I saw it important to refactor our test workflows for 2.x version. Having a bigger picture of having more workflows joining master soon or latter, here are two things I'm putting in consideration:

1. The test workflow name should be able to indicate the module name, as well as the version and what the test accomplishes . For example `refapp2ClinicalVisitTest` from the technical angle one is at least able to know that the workflow is for reference-application and the version is of 2.x and the test is accomplishing a clinical visit user story.
I was not convinced with myself to put `x-`value on the integer version value of the module since this is a command that triggers the test workflow to run and from the technical end refapp2ClinicalVisitTest is good enough to distinguish the same command for 3.x version that would look like refapp3ClinicalVisitTest

2. Being that we have two different versions of RefApp ie 2.x vs 3.x we would not only need to have their respective test workflows in the same place but as well have better names to them. I think having all the test workflows of 2.x RefApp in the same place(as a unit) as well those of 3.x in their place too(as a unit) shouts out well from the technical angle and I think its the better way!

PR #76, opened 2021-08-17
RATEST-194: Create a feature file For Forms Management

sherrif10:
https://issues.openmrs.org/browse/RATEST-194


kdaud:
Any reason for creating feature file for forms ?

sherrif10:
Sure the ideal reason would be automating this form in a new workflow, Any suggestion that would deny this test to be worked upon.

kdaud:
> Any suggestion that would deny this test to be worked upon.

If the test was not covered by the legacy selenium tests then its worth to be worked !

kdaud:
> Any suggestion that would deny this test to be worked upon.

If the test was not covered by the legacy selenium tests then its worth to be worked !

sherrif10:
Wont do since it was covered in the old test framework

PR #79, opened 2021-08-18
RATEST-150: E2E Search and Registration of New Patient

sherrif10:
https://issues.openmrs.org/browse/RATEST-150


HerbertYiga:
thanks @sherrif10  RATEST-150:RATEST-150:  seems to be a repetition on the pr title

sherrif10:
Wondering how these are failing

sherrif10:
Thanks @HerbertYiga

sherrif10:
> Why not rather combine `RegistrationSteps` and `SearchAndRegistrationStpes` into one class ?

This might not work and bring out the use case, i do think we might not continue with this implementation of such work flow

kdaud:
Ideally `RegistrationSteps  `is a subsection of `SearchAndRegistrationSteps`  so combining them into one sounds well and should resolve the issue.

@sherrif10 do you have an idea on how you  to make things resolved rather than the opinion above ?

sherrif10:
FYI  this test is not tended not to touch any class, we are trying to re use method step definition classes which at some point not implementable.  i dont think combining the two have any issue but i would not want to touch these tests, However initially i though of using inheritance approach and res-use of method and things worked perfect for my hook. but still this is leading me to touch more step definition classes and method steps hence returning nullpointer Exception.

kdaud:
As mentioned in the first paragraph [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/79#issuecomment-908459329) combining these two StepDefinition classes into one should be corresponded by having as well one feature file holding all the user stories. This will make the test pass with ease.

sherrif10:
Its actually not about making the test to pass, but also having a meaningfull implementation of the steps and feature files. Combining the two will result into refactoring registrationsteps, registrationsteps, and there feature files. This can be implemented but am seeing it in a perspective of waste of resources.

sherrif10:
i think since we have them already in master , we can pause this, However still am open to hear from different ideas cc @dkayiwa   suggestion whether we can go ahead an merge all this into one workflow.

kdaud:
Have dropped my comment on the [ticket](https://issues.openmrs.org/browse/RATEST-150) with a recommendation.

PR #87, opened 2021-08-23
Fix patient deletion issue in 3.x Clinical flow

jayasanka-sack:
## Purpose
The purpose of this PR is to fix the patient deletion issue in the `3.x Clinical flow`
See: https://issues.openmrs.org/browse/RATEST-201

## Approach
- Pass the UUID instead of the identifier


jayasanka-sack:
@brandones the deletion issue is now fixed. (removed the `purge=true` option in the previous PR which address the issue. however, the identifier was provided instead of the UUID. It was fixed here)

Btw, the tests fail randomly. Most of the time it's because the page doesn't get loaded within the timeout event it has increased to `120` seconds (previously it was `40`).

brandones (approved):
Looks good, thank you @jayasanka-sack !

PR #96, opened 2021-08-31
RATEST-210:Refactor RefApp 3.x registrationsteps to reflect two scenarios

sherrif10:
https://issues.openmrs.org/browse/RATEST-210    
What changed
For this workflow to reflect exactly what it handles, it should be considerate  to be reflecting the exact class, so i refactored registration.js classand its feature file for future use and followup making it easy  cc @jayasanka-sack  @brandones  @kdaud


jayasanka-sack (commented):
Thanks for getting this done @sherrif10 .

How about using `search_and_registration`

PR #102, opened 2021-09-01
RATEST-150R:New E2E Test For Registration Workflow

sherrif10:
https://issues.openmrs.org/browse/RATEST-214

PR #108, opened 2021-09-08
RATEST-150S:New E2E Test For FindPatient Workflow

sherrif10:
https://issues.openmrs.org/browse/RATEST-221

### DIFF
--- a/.github/workflows/refapp-2x-findPatient.yml
+++ b/.github/workflows/refapp-2x-findPatient.yml
@@ -1,0 +1,51 @@
+name: RefApp 2.x Find Patient
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  repository_dispatch:
+    types: [qa]
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    services:
+      firefox:
+        image: selenium/node-firefox:3.141.59-gold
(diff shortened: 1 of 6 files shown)


sherrif10:
Will revert this as soon as possible, am testing out single scenario. am will be reverting this for more tech reviews.Thanks

kdaud:
> Will revert this as soon as possible, am testing out a single scenario.

 @sherrif10 here is my thought about merging and reverting!

 Its not a good ideal to be testing a workflow in master build plan since we have local instances to test the workflow from and Actions to use for online testing purposes for these workflows.

 When a workflow passes with a local instance it can then be pushed to Actions to verify and if it passes the plan then reviewers can take a look at PR before its merged.

Otherwise its not a good practice of merging PRs and reverting commits because of testing purposes. Personally have not seen any senior dev doing what your doing with merge privileges.@sherrif10 when your doing this, do put in mind that those we are mentoring in the team will also do what your doing once they have been accorded with the privilege ? 

@sherrif10 if you have any thing to say, let me know otherwise I would say lets use well the privilege as expected !!

kdaud:
BTW @sherrif10 did get a chance to see  this comment [here](https://github.com/openmrs/openmrs-contrib-qaframework/pull/102#issuecomment-914051052) ?

PR #121, opened 2021-09-14
RATEST-230:Use cucumber Examples Table to fetch Patient name

sherrif10:
https://issues.openmrs.org/browse/RATEST-230  cc @kdaud 
Have re opened this pull request again

### DIFF
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
@@ -43,9 +43,9 @@ public class FindPatientSteps extends Steps {
 		assertNotNull(getElement(By.className("dataTables_empty")));
 	}
 
-	@And("User enters familyName")
-	public void enterPatientName(String familyName) {
-		findPatientPage.enterPatient(testPatient.familyName);
+	@And("User enters patient identifier")
+	public void enterPatientName() {
+		findPatientPage.enterPatient(testPatient.identifier);
 	}
 
 	@Then("Search Page returns patients")
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
@@ -40,9 +40,9 @@ public class FindPatientSteps extends Steps {
 		assertNotNull(getElement(By.className("dataTables_empty")));
 	}
 
-	@And("User enters John Smith")
-	public void enterJohnSmith() {
-		findPatientPage.enterPatient("John Smith");
+	@And("User enters {string} patientName")
+	public void enterPatientName(String patientName) {
+		findPatientPage.enterPatient(patientName);
 	}
 
 	@Then("Search Page returns patients")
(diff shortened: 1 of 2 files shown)


src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
```diff
@@ -40,9 +40,9 @@ public void noPatients() {
 		assertNotNull(getElement(By.className("dataTables_empty")));
 	}
 
-	@And("User enters John Smith")
-	public void enterJohnSmith() {
-		findPatientPage.enterPatient("John Smith");
+	@And("User enters {string} patientName")
+	public void enterPatientName(String patientName) {
```
kdaud, line 44: @sherrif10 why not renaming the method with a genericName that can cater for searching a patient using patientIdentifier or patientName ?

sherrif10:
This  kind of worklow caters for patienName search

kdaud:
The test will be using the method to search for the patient created in the db by the test its self, so its hardly unpredictable to guess the right patient name or patientIdentifier to pass in the method so you may not need to pass a string within the method call. 

The idea would be having something like
```js
public void searchPatient()  {
  findPatientPage.enterPatient(testPatient.identifier); 
  }

```

sherrif10:
> ```js
> public void searchPatient()  {
>   findPatientPage.enterPatient(testPatient.identifier);
> ```

By default this will create patientName = John Smith , so its a two way round

sherrif10:
This implementation comes from findPatientPage defined as its return type as a string  here https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/test/java/org/openmrs/reference/page/FindPatientPage.java#L25. and your method allows an identifier which is an integer, we can probably have another ticket that refactors this kind of implementation, but I do think ,for now we can simply make it the way I has been initially.

sherrif10:
@kdaud  waiting for your reviews

kdaud:
Passing the patient name within the method requires the exact patient to have been created in the db before.

```js
public void enterPatientName("John Smith")  {

  }

```
@sherrif10 the question will be where in the workflow is John Smith be created before his name is passed in the method ?

sherrif10:
> where in the workflow is John Smith created before his name is passed in the method ?

I think what we can do is to also merge this test into registrationPatient test I see the idea you have is that there no independent tests like these never exist. This test is searching for a patient in the db, It only search for a patient already  registered in the database.We are not supposed to create a patient again, Using cucumber table to fetch any given data is generic, cucumber data filters are not specific to the tied data, Rather any data you pass through a string variable is reflected in the function .

sherrif10:
@kdaud  does the last commit pick your approach.

kdaud:
> does the last commit pick your approach.

There are two things am looking into;
1. passing  a patient name in the method enterPatientName("PatientName")   requires knowing the exact name of the patient and this will require creating a patient in the db with the same name before being passed into the method params. 

2. On the other hand finding a patient that is randomly created by (`testPatient = createTestPatient()`) may not need to be passed in the method params  since the names given to the patient are randomly selected.  With this approach we can have something like

 ```js
public void searchPatient()  {
   findPatientPage.enterPatient(testPatient.identifier);
  }

```

sherrif10:
Thanks @kdaud , i will be closing this ticket and we shall handle this improvements in this ticket https://github.com/openmrs/openmrs-contrib-qaframework/pull/127/files

PR #122, opened 2021-09-15
RATEST-232: Failure when validating the vital value of respiratory rate

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-232)

Description -> Failure when validating the vital value of respiratory rate

PR #126, opened 2021-09-15
RATEST-233: Make local instance have the same versions of modules as qa-refapp

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-233)

Description ->  Make local instance have the same versions of modules as qa-refapp


kdaud:
Note:
This PR can not pass Actions because of the bug that is being fixed in this [PR](https://github.com/openmrs/openmrs-contrib-qaframework/pull/122) which has not yet been merged into master. Either of these PRs can not pass Actions until one of them get merged.

PR #140, opened 2021-10-04
RATEST-244:Update chrome driver to  94.0.4606.71

sherrif10:
https://issues.openmrs.org/browse/RATEST-244

PR #144, opened 2021-10-07
Revert "RATEST-244:Update chrome driver to  94.0.4606.71"

dkayiwa:
Reverts openmrs/openmrs-contrib-qaframework#140


mozzy11:
hello @dkayiwa , did this upgrade cause any build failure some where ??
its what i currently use for the isante and OpenELIS qa-frameworks.
Anyways am not sure exactly about the chrome what Sharrif used ,but the upgrade would be fine

PR #180, opened 2021-10-22
RATEST-259: Format qaframework module for quality assurance

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-259)

Description -> Format qaframework module for quality assurance.

Background
When running any local build plan, cucumber-selenium auto-formats the code base resulting into undesirable format which needs to be overridden to ensue code sanity within the repo.

### DIFF
--- a/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java
@@ -156,8 +156,7 @@ public class TestBase implements SauceOnDemandSessionIdProvider {
 
 			capabilities.setCapability("name", testMethod);
 
-			capabilities.setCapability("commandTimeout",
-					MAX_SAUCELAB_COMMAND_TIMEOUT_IN_SECONDS);
+			capabilities.setCapability("commandTimeout", MAX_SAUCELAB_COMMAND_TIMEOUT_IN_SECONDS);
 
 			String buildNumber = System.getProperty("buildNumber");
 			if (!StringUtils.isBlank(buildNumber)) {
@@ -166,8 +165,7 @@ public class TestBase implements SauceOnDemandSessionIdProvider {
 
 			String saucelabsTunnel = System.getProperty("saucelabsTunnel");
 			if (!StringUtils.isBlank(saucelabsTunnel)) {
-				capabilities
-						.setCapability("tunnel-identifier", saucelabsTunnel);
+				capabilities.setCapability("tunnel-identifier", saucelabsTunnel);
 			}
 
 			String branch = System.getProperty("branch");
@@ -175,10 +173,8 @@ public class TestBase implements SauceOnDemandSessionIdProvider {
 				capabilities.setCapability("tags", branch);
 			}
 
-			if (TestProperties.DEFAULT_WEBDRIVER
-					.equals(properties.getBrowser())) {
-				capabilities.setCapability(
-						CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR,
+			if (TestProperties.DEFAULT_WEBDRIVER.equals(properties.getBrowser())) {
+				capabilities.setCapability(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR,
 						UnexpectedAlertBehaviour.IGNORE);
 			}
 
@@ -187,14 +183,11 @@ public class TestBase implements SauceOnDemandSessionIdProvider {
 					+ sauceLabsAuthentication.getAccessKey() + "@"
 					+ sauceLabsHubUrl + "/wd/hub"), capabilities);
 
-			this.sessionId = (((RemoteWebDriver) driver).getSessionId())
-					.toString();
-			System.out.println("Running " + testMethod
-					+ " at https://saucelabs.com/tests/" + this.sessionId);
+			this.sessionId = (((RemoteWebDriver) driver).getSessionId()).toString();
+			System.out.println("Running " + testMethod + " at https://saucelabs.com/tests/" + this.sessionId);
 		} else {
 			System.out.println("Running locally...");
-			final TestProperties.WebDriverType webDriverType = properties
-					.getWebDriver();
+			final TestProperties.WebDriverType webDriverType = properties.getWebDriver();
 			switch (webDriverType) {
 				case chrome :
 					driver = setupChromeDriver();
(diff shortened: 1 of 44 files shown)


kdaud:
@sherrif10 any recommendation on the comment [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/180#issuecomment-953773735)?

kdaud:
@Irenyak1 @Ndacyayisenga-droid the commit here will solve the problem you are facing while running any local build plans on the tests you are working on. 

@sherrif10 have you had a look at my opinion [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/180#issuecomment-953773735)?

Irenyak1:
> @Irenyak1 @Ndacyayisenga-droid @mherman22 the commit here will solve the problem you are facing while running any local build plans on the tests you are working on.
> 
> @sherrif10 have you had a look at my opinion [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/180#issuecomment-953773735)?

@kdaud , That will be a great break through because the developer has had painful moment due to auto formatting of changes from upstream especially when a test is run locally.

mherman22:
> I think this format is most likely to be coming from cucumber-selenium plugin that formats things this way whenever a build plan is made ie (the code format in this PR was the one existing in the module before this [commit](https://github.com/openmrs/openmrs-contrib-qaframework/commit/14f8dc0893635e145eaa802fc923755fd3c4d8a7) was merged). Have not investigated on how this formatter can be overridden, otherwise have started observing some PRs where things are messed up during committing changed files.
> 
> I suggest we have this guy merged as investigation takes place on how the selenium-cucumber format can be overridden otherwise its quite hard inspecting over 64 files whose changes are brought about by any local build plan.

ooh that's great, thanks for pointing that out. i have tried numerous times t avoid the formatter but to no avail. i think it would be of great addition if formats done here as you suggested

sherrif10:
@kdaud  can we first resolve clinical visit workflow which is still breaking ?

kdaud:
> can we first resolve clinical visit workflow which is still breaking ?

I resolved the test breakage in this [commit](https://github.com/openmrs/openmrs-contrib-qaframework/commit/d36b24b8caaa7a8f6c5fd926a1e69794e83a135d) though the test its self has caught about a bug as per the talk post [here](https://talk.openmrs.org/t/automated-test-for-booking-an-appointment-catches-a-bug/34993)( this is one of the advantages of automated tests that the community is benefiting from).

According to @dkayiwa comment [here](https://talk.openmrs.org/t/build-failure-for-trunk-core2-15-plan/34872/102) the bug is caused as a result of switching platform 2.4.0 to 2.5.0.alpha3 and  currently being addressed by the platform team.

The temporary solution for now if we want the test to pass would be switch the automated tests to spin from demo.openmrs.org. @sherrif10 any suggestion? otherwise we need to relieve pain to the squad members facing issues being addressed in this PR!

kdaud:
@sherrif10 what's the way forward, are we switching both github actions and Bamboo to temporary spin from the demo server? But I think its better to leave Bamboo track that bug and only switch github actions to spin from the demo server! What your opinion?

dkayiwa:
@kdaud no need to switch because it was a very simple fix.

kdaud:
Thanks @dkayiwa for the follow up on the bug.

kdaud:
@sherrif10 waiting for your recommendation on this PR!

sherrif10:
@kdaud , Have you tried used this maven configuration 

> <dependency>
>   <groupId>com.google.googlejavaformat</groupId>
>   <artifactId>google-java-format</artifactId>
>   <version>1.12.0</version>
> </dependency>

sherrif10:
@dkayiwa  @ibacher  is there any configuration for formatting code by default. Have tried to use different format configurations however am not certain it works as expected.

kdaud:
> Have you tried used this maven configuration

Not really!

does it solve the issue?

sherrif10:
@kdaud  you could consume merge conflicts.

kdaud:
> @kdaud you could consume merge conflicts.

done the needful !!

kdaud:
@sherrif10 have you had a look again on this PR?

src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java
```diff
@@ -103,13 +103,19 @@ public void failed(Throwable t, Description test) {
...
-		String sauceLabsAccessKey = testProperties.getProperty("SAUCELABS_ACCESSKEY", null);
-		sauceLabsHubUrl = testProperties.getProperty("saucelabs.hub.url","ondemand.saucelabs.com:80");
-
-		if (!StringUtils.isBlank(sauceLabsUsername)&& !StringUtils.isBlank(sauceLabsAccessKey)) {
-			sauceLabsAuthentication = new SauceOnDemandAuthentication(sauceLabsUsername, sauceLabsAccessKey);
-			sauceLabsResultReportingTestWatcher = new SauceOnDemandTestWatcher(this, sauceLabsAuthentication);
+		String sauceLabsUsername = testProperties.getProperty(
+				"SAUCELABS_USERNAME", null);
+		String sauceLabsAccessKey = testProperties.getProperty(
+				"SAUCELABS_ACCESSKEY", null);
```
sherrif10, line 109: Yes @kdaud  as you can see this is not the best we can let go, because am still wondering if someone reviews this code and it looks like this, it would not show quality code. i don't know whether you understand me. Is it fine if we can try out another option of making our code look more cleaner than this, i believe we can find out another option of making our code look more proffessional.What is your suggestion.

kdaud: >Is it fine if we can try out another option of making our code look more cleaner than this, i believe we can find out another option of making our code look more proffessional

It may require some investigation on how this can be achieved of which am not have time for now.

>i don't know whether you understand me

yes i get point home.

Well, as of now the contributors are facing a challenge of the module being formatted on any local build plan and makes existing files tracked as changed files and have also realized some of them are not committing their changes as a result of the issue.

@sherrif10 if you have time, you can get the best solution to the issue otherwise I had thought this to be a temporarily solution to relieve the contributors pain as will get time in the near by future and investigate to find out the solution.

sherrif10: Thanks @kdaud  for understanding the point, we need clean code. Ok let me investigate hope to come up with something, Meanwhile lets have final discussion of writing automated test in new bdd approach from distro.

kdaud:
The solution here is sufficient enough to ensure code sanity and getting rid of the problem that has been occurring whenever a local build plan is made on the repo.

@sherrif10 most of the changes are ensuring code sanity but the most important one that has solved  the issue was to get rid of the java-formater plugin that was causing resulting into the undesirable code format whenever a local build plan is made on the repo.
```js
 <plugin>
                <groupId>com.googlecode.maven-java-formatter-plugin</groupId>
                <artifactId>maven-java-formatter-plugin</artifactId>
                <version>${formatter.version}</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>format</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

kdaud:
@sherrif10 have you had a look at the solution in this commit regarding code sanity?

sherrif10:
Is there any configuration you have done on your IDE

kdaud:
> Is there any configuration you have done on your IDE

no

The solution is referenced in the second paragraph of the above [comment](https://github.com/openmrs/openmrs-contrib-qaframework/pull/180#issuecomment-959924569)!

sherrif10 (approved):
LGTM . Am keeping an eye on this change immediately it gets into master.

kdaud:
@dkayiwa can do the needful concerning this commit!

PR #201, opened 2021-11-01
RATEST-258:Migrate All The Tests From Distro Into qa Framework

sherrif10:


sherrif10:
cc @dkayiwa  @ibacher  Am kindly requesting for your help on this because am not cleary understanding why drivers are returning null while before migration of tests they fail

sherrif10:
@dkayiwa  Yes i tried seeking help and i think he tried out same thing to confirm.

sherrif10:
@dkayiwa  . Thanks here are steps

1. Copy and paste any selenium test from distro . Forexample  [startActiveTest](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/test/java/org/openmrs/reference/StartVisitTest.java)
2. Import[ LocationSensitiveApplicationTestBase.java](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/aa9aec9bf173af9aecb371143b4f7a96a7a40d58/ui-tests/src/test/java/org/openmrs/reference/https://github.com/openmrs/openmrs-distro-referenceapplication/blob/aa9aec9bf173af9aecb371143b4f7a96a7a40d58/ui-tests/src/test/java/org/openmrs/reference/LocationSensitiveApplicationTestBase.java#L11#L11)class since its parent class for startVisitTest
3. Then we change all the import statements to import from import org.openmrs.contrib.qaframework.
4. From here we can run our test as junit test.
5

ibacher:
@sherrif10 So, let's walk through this a bit:

1. The tests are consistently failing with an NPE at [this line](https://github.com/openmrs/openmrs-contrib-qaframework/pull/201/files#diff-62afb9c7433e51cde733bf88a8f3caefbcf157e06124bece662f9c16bf288fe8R88). If this line is throwing an NPE it's because `parent` is null so that trying to access `parent.driver` throws the NPE.
2. From the stacktraces, I can see that we're consitently calling the constructor that's throwing the NPE with a line [like this one](https://github.com/openmrs/openmrs-contrib-qaframework/pull/201/files#diff-c34fff6eea005d89d2d0f5635e75ad6d0893f22022bc69807c4c1aad4437cc8fR44) where we call `new HomePage(page)` in a class that is a sub-class of [`TestBase`](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java). So this means that that the `page` property in `TestBase` is `null` at this point.

I notice that in `TestBase` the `page` property is not initially populated with a value. This leads to two questions:

1. Where is the `page` property supposed to get a non-null value from?
2. Is there a reason why the part of the code that is supposed to populate the `page` property would not be being called?

sherrif10:
> 2\. From the stacktraces, I can see that we're consitently calling the constructor that's throwing the NPE with a line [like this one](https://github.com/openmrs/openmrs-contrib-qaframework/pull/201/files#diff-c34fff6eea005d89d2d0f5635e75ad6d0893f22022bc69807c4c1aad4437cc8fR44) where we call `new HomePage(page)` in a class that is a sub-class of [`TestBase`](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java). So this means that that the `page` property in `TestBase` is `null` at this point.

This is very true page property in testBase is null

sherrif10:
Thanks @dkayiwa @ibacher  i think if we reach at the point where page is null picking it from TestBase class, we might get the solution, Again shall we also try to change ### homePage = new HomePage(page) in before method and do it plainly though this already comes when page is null from testBase

kdaud:
The stacktrace  thrown suggests this is a problem with `driver` variable not initialized. Probably its because the driver is failing to be inherited from the parent(base) class. The program tries to use the class's variable instead of the super variable but the class variable is not initialized, the constructor just calls to the super constructor hence throwing null pointer exception.

mozzy11:
this property [login.auto](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/src/test/resources/org/openmrs/uitestframework/test.properties#L6) is set to false 
and and the [page object](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java#L212)  is not getting instatiated because of this [check](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/src/test/java/org/openmrs/contrib/qaframework/helper/TestBase.java#L210)

sherrif10:
Thanks @mozzy11  for looking into this.

sherrif10:
Let me open a new ticket

PR #217, opened 2021-11-09
RATEST: Migrate selenium legacy tests into qaframework module

kdaud:


sherrif10:
This is nice approach, However to add bit of comment, when these tests increase in number over 80 tests, driver will take long to spin up app the tests at once especially  for the tests that run parallel and those simultaneously(selenium tests), its too tricky now to come in sense of refactoring all the elements to increase the capability of the driver to run test scripts smoothly

sherrif10:
Feel free to add morethan 6 selenium tests and see the magic of github actions.

PR #243, opened 2021-11-24
Adding ui tests as submodule into qaframework

sherrif10:
cc @kdaud  @dkayiwa  @ibacher  . Hade been trying to use this approach but am not sure where i could be messing up

However still it seems something am trying to achieve brings me back to the triggering of RUn test class. COuld you please help me understand it. Your guidance will be of help.


kdaud:
@sherrif10 how about we just add a workflow that triggers selenium legacy tests ?

sherrif10:
> how about we just add a workflow that triggers selenium legacy tests ?

This is the workflow that triggers the selenium tests https://github.com/openmrs/openmrs-contrib-qaframework/runs/4320604745?check_suite_focus=true    . However this wont work because this wont have a folder named ui-tests where selenium tests will be found.

sherrif10:
Again,The idea is to have get away of how we are triggering it locally first, The challenge is that this is trigger npm script with in RunTest class which by default triggers all tests which we dont want.

kdaud:
> The idea is to have get away of how we are triggering it locally first

@sherrif10 let me work around the ideal and probably come up with a working PR before the day ends so that we have this deal done.

sherrif10:
That's sounds good deal, am also working around it as you also come up

PR #246, opened 2021-11-29
RATEST: Remove the formatter version since the plugin no longer exists

kdaud:
Goal:
Get rid of the formatter version since the plugin no longer exists


sherrif10 (approved):
Thanks @kdaud

PR #247, opened 2021-11-30
RATEST: Migrate selenium legacy tests into qaframework

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-258)

Goal -> Migrate selenium legacy tests into qaframework


dkayiwa:
@kdaud great to see you resume working on this! 👍

sherrif10:
Sure

sherrif10:
Wow this is a great achievement, since i had same code though not pushed, am closing it an leaving you accomplishing it. Otherwise well done. @kdaud

kdaud:
> Does this workflow triggers any single selenium test ?

@sherrif10 do you mind phrasing the question differently so as to give appropriate response on the subject?

kdaud:
> Does this workflow triggers any single selenium test ?

@sherrif10 do you mind phrasing the question differently? so that I respond appropriately!

.github/workflows/refapp-2x-legacy.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: run db and web containers
+        run: |
+          docker-compose -f docker/docker-compose-refqa.yml up -d
+      - name: wait for openmrs instance to start
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 1; done
+      - name: Run qaframework on selenium
+        run: |
+          mvn clean install -DskipTests=true
+          cp -f src/test/resources/org/openmrs/uitestframework/test-local.properties src/test/resources/org/openmrs/uitestframework/test.properties
+          cd src/main/java
```
sherrif10, line 37: Thanks @kdaud , i meant, is this workflow triggering selenium tests specifically.

kdaud: `cd src/main/java` navigates the folder where the selenium legacy tests are residing and the command that is triggered does the building of the tests!

sherrif10: Are we supposed to see a number of built selenium  tests with in the logs or the workflow is supposed to only run BDD written  tests ?

sherrif10: According to this logs [here](https://github.com/openmrs/openmrs-contrib-qaframework/runs/4375756745?check_suite_focus=true)  , its showing still runTest class gets triggered with default workflows like  **refappSelenium** and i think this is what i was also trying to overcome, Am glad you are into this too and great work man.

README.md
```diff
@@ -24,7 +24,7 @@ Install & Upgrade Tests |
 [All Chrome Tests](https://ci.openmrs.org/browse/CONTRIB-QA) ![Build Status](https://ci.openmrs.org/plugins/servlet/wittified/build-status/CONTRIB-QA)
 [![All Firefox Tests](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/qa.yml/badge.svg?branch=master)](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/qa.yml)
 [![RefApp 2.x Clinical Visit](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/refapp-2x-clinical-visit.yml/badge.svg)](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/refapp-2x-clinical-visit.yml)
-[![RefApp 2.x Patient Condition](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/refapp-2.x-condition.yml/badge.svg)](https://github.com/openmrs/openmrs-contrib-qaframework/actions/workflows/refapp-2.x-condition.yml)
```
sherrif10, line 27: This change looks okay  however it would be awesome to commit only whats required otherwise we might loose track in future

package.json
```diff
@@ -25,12 +25,12 @@
...
     "microfrontend": "cypress-tags run -eTAGS='@cypress and @microfrontend'",
-    "refapp3Login": "cypress run --spec src/test/resources/features/refapp-3.x/01-login/login.feature",
-    "refapp3PatientRegistration": "cypress run --spec src/test/resources/features/refapp-3.x/02-patient-registration/patient-registration.feature",
-    "refapp3PatientSearch": "cypress run --spec src/test/resources/features/refapp-3.x/06-patient-search/patient-search.feature",
-    "refapp3Settings": "cypress run --spec src/test/resources/features/refapp-3.x/03-settings/settings.feature",
-    "refapp3ClinicalVisit": "cypress run --spec src/test/resources/features/refapp-3.x/04-clinical-visit/clinical-visit.feature",
-    "refapp3VitalsAndTriage": "cypress run --spec src/test/resources/features/refapp-3.x/05-vitals-and-triage/vitals-and-triage.feature"
+    "refapp3Login": "cypress run --spec src/test/resources/refapp-3.x-features/refapp-3.x/01-login/login.feature",
+    "refapp3PatientRegistration": "cypress run --spec src/test/resources/refapp-3.x-features/refapp-3.x/02-patient-registration/patient-registration.feature",
```
sherrif10, line 29: Same thing here we could track this change in separate pull request

package.json
```diff
@@ -25,12 +25,12 @@
     "coreUpgrade": "mvn test -Dcucumber.filter.tags='@selenium and @initialSetup and @upgrade'",
     "ocl": "./node_modules/.bin/cypress-tags run -eTAGS='@cypress and @ocl'",
     "microfrontend": "cypress-tags run -eTAGS='@cypress and @microfrontend'",
-    "refapp3Login": "cypress run --spec src/test/resources/features/refapp-3.x/01-login/login.feature",
```
sherrif10, line 28: Any reason for changing the feature folders here

.github/workflows/refapp-2x-legacy.yml
kdaud: @sherrif10 what is your thought?

sherrif10: As long as we allow the selenium tests gets triggered independently , i think we will be ready to go. we need to override runTest class not to be triggered since it comes along with npm scripts which we might not want with selenium tests.

package.json
kdaud: The refapp-3.x features have been removed from features folder  to avoid [cucumberOptions()](https://github.com/openmrs/openmrs-contrib-qaframework/blob/dc522ce353fe9c9f53ae01acdf72420aa35b7187/src/test/java/org/openmrs/contrib/qaframework/RunTest.java#L20) not to track them as unimplemented. And thus making a chance in the scripts that locaates the respective feature files.

.github/workflows/refapp-2x-legacy.yml
kdaud: Most likely we need to trigger the selenium legacy tests via `mvn clean <params>`!

package.json
sherrif10: Was it failing initially ?

.github/workflows/refapp-2x-legacy.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Checkout qaframework
+        uses: actions/checkout@master
+        with:
+          repository: ${{github.repository}}
+      - name: run db and web containers
+        run: |
+          docker-compose -f docker/docker-compose-refqa.yml up -d
+      - name: wait for openmrs instance to start
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 1; done
+      - name: Run qaframework on selenium
```
sherrif10, line 33: @kdaud  still am not understanding what this workflow triggers, to me it seems to be the working as this [workflow ](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/.github/workflows/qa.yml), is there a difference. Are we really triggering selenium or we can simply handle them without triggering them as along as they dont contradict, I mean simply doing the way you have done but without this workflow because it seems not to be working as expected@kdaud  still am not understanding what this workflow triggers, to me it seems to be the working as this [workflow ](https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/.github/workflows/qa.yml), is there a difference. Are we really triggering selenium or we can simply handle them without triggering them as along as they dont contradict, I mean simply doing the way you have done but without this workflow because it seems not to be working as expected

sherrif10:
For selenium tests to be triggered separately i think we should create a submodule that only contains those selenium tests have you tried this out ?

.github/workflows/refapp-2x-legacy.yml
kdaud: @sherrif10 I agree!

kdaud: Now that we have the BDD workflow running independently as we intended them to be running!

sherrif10:
@kdaud  any response to the above comment

kdaud:
> For selenium tests to be triggered separately i think we should create a submodule that only contains those selenium tests have you tried this out!

Good idea @sherrif10 !
Now that the BDD tests are running independently off selenium, we can ensure selenium legacy tests are also triggered separately. Currently the tests are residing in the module but not running on any ci plan.

We would have them reside the way they are now if transforming them into BDD framework was the next on our plan so as to have their correspondant parent modules(where there have migrated) are deleted to avoid duplicate.

But I think the better way is to have them running ci as well so that they continue testing their respective functionalities than waiting to be transformed into BDD tests.

@sherrif10 what is your thought?

sherrif10:
> Currently the tests are residing in the module but not running on any ci plan.

i think initially  the idea was to have them also being triggered by the ci.

sherrif10:
> But I think the better way is to have them running ci as well so that they continue testing their respective functionalities than waiting to be transformed into BDD tests then do the work.

i agree with you, Have you tried out creating a submodule and see if that works probably i might had missed something.

kdaud:
> For selenium tests to be triggered separately i think we should create a submodule that only contains those selenium tests have you tried this out ?

Have given an attempt on creating a submodule but seems not to work neither. From observation, seems like we may not be able to have two test frameworks in one repository running together and have not known of a test framework architecture that can support more than one automation framework in a single repo! 

For now I would go back to the option of having the selenium legacy tests dwell in the repo without running as transformation into BDD takes place. 

@sherrif10 what is your thought?

sherrif10:
@kdaud incase of help or suggestion, Feel free to reach out, if all the ideas dont work out, we can seek more help , but personally i would propose to do everything in new bdd approach since its the future way.

kdaud:
@sherrif10 I am done with the deal for migrating the selenium legacy tests and waiting for your approval.

As per this commit, we have
1. BDD workflow tests running independently without trigering the selenium legacy tests in any way
2. Selenium legacy tests have thair workflow that triggers them during ci.

> i would propose to do everything in new bdd approach since its the future way.

Yes I know this will be a process and now once this commit is merged, we can have those duplicate files in other repos(distro and uitestframework) removed since you made pull requests already. Then when the team is ready to dive into transforming the selenium tests into BDD tests, then the process can kick off given that our Team is growing faster so its an advantage we can leverage-on.

.github/workflows/refapp-2x-legacy.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          mvn clean install -DskipTests=true
+          cp -f src/test/resources/org/openmrs/uitestframework/test-local.properties src/test/resources/org/openmrs/uitestframework/test.properties
+          cd src/legacy/java
```
sherrif10, line 37: This is amazing effort @kdaud , One more slight question, does is this folder checked out during the build instantiation ?

kdaud: With ci, the folder is checkout by that command line `cd src/legacy/java`. But locally one has to do it manually!

kdaud: BDD tests does not trigger the selenium legacy tests since their workflow does not check out that specific folder.

sherrif10: > 

Is that workflow neccessary since it doesnt trigger anything, what happens if its not there

sherrif10: > 

Is that workflow necessary since it doesn't trigger anything, what happens if its not there

kdaud:
@sherrif10 I stepped back to look into this PR in regard to the review [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/247#discussion_r762445484) and I discovered the selenium legacy tests were not being triggered. 

I found out for selenium legacy tests to be triggered independently we needed to create sub modules and that's what have implemented(creating two modules: `qaframework-bdd-tests` & `qaframework-legacy-tests`). So as per this PR we have 
1. BDD tests running independently when triggered and also
2. Selenium legacy tests running independently when triggered.

However, I encountered  a scenario where every module has to host pages which makes them duplicate in both modules. Do you know of a way we can achieve having them not duplicated but being able to be used in both modules?

kdaud:
One more light on the subject [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/247#issuecomment-986441182) is that;
 For each test (either BDD test or Selenium legacy test), dependencies need to build first before the test/s is run. And building the dependencies required the test to have navigated into the sub-module since these tests dependencies vary. Now by removing page files in one of the sub-module means during the dependency build up, those page files will be missing and result into PageNotFoundException.

#sherrif10  do you have an idea on how to achieve the intended goal so that these page files are not duplicated in the repo?

sherrif10:
@kdaud. Thanks for the initiative. Like the fact that you have invested
time into this. Here is my question, are the tests need pages to be into
separate folder to run, I do think it doesn't necessarily where pages stay,
the problem is with where we direct our run folder class

kdaud:
> @kdaud. Thanks for the initiative. Like the fact that you have invested time into this. Here is my question, are the tests need pages to be into separate folder to run, I do think it doesn't necessarily where pages stay, the problem is with where we direct our run folder class

> Here is my question, are the tests need pages to be into separate folder to run

Not really we can have them in one folder, However, I think we would refactor them from `src/test/java` folder for a better architecture. Then we can the pages stored in `src/main/java `  folder. I found REST WS [module](https://github.com/openmrs/openmrs-module-webservices.rest) test architecture a good one which we can use as well in qaframework to separate folders for tests files and pages. As per its architecture the resources are kept in a separate [folder](https://github.com/openmrs/openmrs-module-webservices.rest/tree/master/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8) from that of the [tests](https://github.com/openmrs/openmrs-module-webservices.rest/tree/master/omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_8). and I find this approach a better one to have in our qaframework module.

> the problem is with where we direct our run folder class

@sherrif10 do you have an idea on this so as to avoid duplicates?

kdaud:
> @sherrif10 do you have an idea on this so as to avoid duplicates?

This [link](https://stackoverflow.com/questions/17492622/specify-common-resources-in-a-multi-module-maven-project) has the idea to meet the requirements.

kdaud:
@sherrif10 do you any further review on this pull before Daniel is invited to wrap it all and have the deal done?

sherrif10:
@kdaud  Thanks for this work. The idea for this ticket was because of the limited time to migrate all the tests. Personally, I think we can start on migrating of these tests into new Bdd since we seem to be having again to migrate them. I currently don't have any other option since this approach would not bring it out very well in a clear context unless. However this would be a good deal to start work on these tests slowly by slowly good enough we have a core team unless  @dkayiwa  have final conclusion..

kdaud:
> I currently don't have any other option since this approach would not bring it out very well in a clear context unless.

As per this pull request we now have;
1. BDD tests running independently when triggered by their respective workflows, and also
2. Selenium legacy tests running independently when [triggered](https://github.com/openmrs/openmrs-contrib-qaframework/blob/f192888a9a76596fd190ee65127729b4e4f908f4/.github/workflows/refapp-2x-legacy.yml#L37-L38)

isn't the objectives above we wanted to achieve for [RATEST-258](https://issues.openmrs.org/browse/RATEST-258)?

sherrif10:
It's not insufficient  for me still , we better spend much time on
migrating all the tests . However I leave decession with @dkayiwa.

dkayiwa:
> we better spend much time on migrating all the tests

Aren't all the tests migrated in this pull request?

sherrif10:
> Aren't all the tests migrated in this pull request?

My bad, Meant migrating all the selenium tests into new BDD, am seeing this as temporary work.

dkayiwa:
Conversion into BDD tests should be a separate stream of work.

sherrif10:
@kdaud  Thanks alot for this awesome work .@dkayiwa  i dont if you have went through the architecture of the current status of the module whether it meets the standards of clean quality assurance framework. Alot of duplicates , some pages have been duplicated from with in separate folders. It looks have achieved the goal but am not sure if really this could be the right approach.

This is going to make work more un balancing especially when we will be using different packages. @kdaud  What is your suggestion of the current status
.

sherrif10:
Forexample , https://github.com/openmrs/openmrs-contrib-qaframework/tree/master/qaframework-bdd-tests/src/main/java/org/openmrs/contrib/qaframework    and this  https://github.com/openmrs/openmrs-contrib-qaframework/tree/master/src/test/java/org/openmrs/contrib/qaframework  have same things ,this is somewhat duplicates .am so exhusiastic about @kdaud  workflow. Alot changed with in this.

dkayiwa:
Thanks @sherrif10 for bringing this up.

@kdaud what do you think of this?

kdaud:
Thanks @sherrif10 for identifying this early enough. Have stepped back and identified the issue as well, and I agree with you to avoid duplicates in the module.

> @kdaud what do you think of this?

@dkayiwa have just made a follow up commit in pull/265 that fixes the issue.

kdaud:
> @kdaud is the followup commit about ignoring failing tests or removing duplicate files?

initially had combined both but have done the needful after taking a second look at https://opensource.com/article/18/6/anatomy-perfect-pull-request

PR #254, opened 2021-12-06
TEST: Watch the behavior of bdd and legacy tests via actions

kdaud:


sherrif10:
The easiest way is to have these tests written in bdd approach , and i guess we should start this milestone as possible , How do you see .

PR #256, opened 2021-12-06
RATEST-277: Replace assertText() with assertPage() to improve the tests run

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-277)

Description -> Replace assertText() with assertPage() to improve the test run.

### DIFF
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
@@ -9,7 +9,7 @@
  */
 package org.openmrs.contrib.qaframework.automation;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 import org.openmrs.contrib.qaframework.RunTest;
@@ -42,13 +42,13 @@ public class AllergiesSteps extends Steps {
 
 	@Then("the system loads Allergies page")
 	public void systemLoadsAllergiesPage() {
-		assertEquals(getElement(patientHeaderId).getText(),getElement(patientHeaderId).getText());
-		assertTrue(textExists("Allergies"));
+		assertPage(allergyPage.waitForPage());
 	}
 
 	@And("a user clicks No Known Allergy button")
 	public void addNoKnownAllergy() {
 		allergyPage.addNoKnownAllergy();
+		allergyPage.waitForPage();
 	}
 
 	@Then("the system add no known allergies into the allergies table")
@@ -59,6 +59,7 @@ public class AllergiesSteps extends Steps {
 	@And("a user clicks Remove No Known Allergy icon")
 	public void removeNoKnownAllergy() {
 		allergyPage.removeNoKnownAllergy();
+		allergyPage.waitForPage();
 	}
 
 	@Then("the system displays unknown in the allergies table")
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
@@ -64,7 +64,7 @@ public class ConditionsSteps extends Steps {
 
 	@Then("System returns to patient dashboard")
 	public void returnToDashboard() {
-		assertPage(dashboardPage.waitForPage());
+		matchPatientIds(patientDashboardId);
 	}
 
 	@And("User clicks on Add new condition")
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
@@ -54,7 +54,7 @@ public class ConditionsSteps extends Steps {
 
 	@Then("System loads Manage Conditions Page")
 	public void systemLoadsManageConditions() {
-		assertNotNull(getElement(addNewCondition));
+		assertPage(conditionsPage.waitForPage());
 	}
 
 	@And("User clicks on Return")
@@ -64,7 +64,7 @@ public class ConditionsSteps extends Steps {
 
 	@Then("System returns to patient dashboard")
 	public void returnToDashboard() {
-		matchPatientIds(patientDashboardId);
+		assertPage(dashboardPage.waitForPage());
 	}
 
 	@And("User clicks on Add new condition")
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/PageRejectedException.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/PageRejectedException.java
@@ -27,5 +27,4 @@ public class PageRejectedException extends RuntimeException {
 			boolean enableSuppression, boolean writableStackTrace) {
 		super(message, cause, enableSuppression, writableStackTrace);
 	}
-
 }
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
@@ -71,6 +71,6 @@ public class FindPatientSteps extends Steps {
 
 	@Then("System loads patient dashboard")
 	public void loadPatientDashboard() {
-		matchPatientIds(firstPatientIdentifier);
+		assertPage(dashboardPage.waitForPage());
 	}
 }
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/FindPatientSteps.java
@@ -71,6 +71,6 @@ public class FindPatientSteps extends Steps {
 
 	@Then("System loads patient dashboard")
 	public void loadPatientDashboard() {
-		assertPage(dashboardPage.waitForPage());
+		matchPatientIds(firstPatientIdentifier);
 	}
 }
--- a/src/test/java/org/openmrs/contrib/qaframework/automation/InpatientSteps.java
+++ b/src/test/java/org/openmrs/contrib/qaframework/automation/InpatientSteps.java
@@ -46,7 +46,7 @@ public class InpatientSteps extends Steps {
 
 	@Then("the system loads find patient record page")
 	public void systemLoadsFindPatientRecordPage() {
-		assertTrue(textExists("Find Patient Record"));
+		assertPage(findPatientPage.waitForPage());
 	}
 
 	@When("a user searches an existing patient from find patient record page")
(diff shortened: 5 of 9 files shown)


src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
```diff
@@ -42,27 +42,28 @@ public void loadAllergiesPage() {
...
 
 	@And("a user clicks No Known Allergy button")
 	public void addNoKnownAllergy() {
 		allergyPage.addNoKnownAllergy();
+		allergyPage.waitForPage();
 	}
 
 	@Then("the system add no known allergies into the allergies table")
 	public void systemAddsNoKnownAllergies() {
-		assertTrue(textExists("No known allergies"));
```
sherrif10, line 56: Havent seen a good reason for this refactor, Do you mind giving me a brief way why this change.

kdaud: Have been noticing random breakage of these automated tests on ci and opted to improve them as a way of ensuring they are stable and can serve the intended purpose otherwise if these tests are not stable then it will have less value as far as our vision is concerned. We need to have them stabilize and reliable on every build plan run !

@dkayiwa  emphasized this via slack where he said `this instability will make it unreliable to be of value`.

kdaud: > Havent seen a good reason for this refactor

assertPage() is more reliable than assertText() in reference to automated tests.

kdaud: Have been noticing random breakage of these automated tests on ci and opted to improve them as a way of ensuring they are stable and can serve the intended purpose otherwise if these tests are not stable then it will have less value as far as our vision is concerned. We need to have them stabilize and reliable on every build plan run !

@dkayiwa  emphasized this via slack where he said `this instability will make it unreliable to be of value`.

sherrif10:
@kdaud  does the invisibility of an element comes after page loads or elements

kdaud:
> does the invisibility of an element comes after page loads or elements

I seem not to get your question well, do you mind phrasing it differently?

kdaud:
@sherrif10 have taken a look at the pardon [above](https://github.com/openmrs/openmrs-contrib-qaframework/pull/256#issuecomment-986909772)?

src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
```diff
@@ -64,7 +64,7 @@ public void clickReturn() {
 
 	@Then("System returns to patient dashboard")
 	public void returnToDashboard() {
-		matchPatientIds(patientDashboardId);
+		assertPage(dashboardPage.waitForPage());
```
sherrif10, line 67: So ideally, let's use this example, suppose the system doesn't return to the dashboard page, and your assertion is waiting for dashboard page instead of a functional  method that directly checks on the known element or id , how would you include that assertion

kdaud: When the system doesn't return to the dashboard page the test must fail the run because the run is expected to be on the dashboard page for the next operation to be successful. As well the previous assertion is good enough to assert the return of the system on the dashboard  since it asserts the patientDashboardId instead of a text. In both assertion if the the system fails the to return on the dashboard page, the test must fail!

kdaud:
@sherrif10 any more reviews to address?

src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
sherrif10: Great, so ideally you are saying the best approach would assertPage instead of text on the page, i do believe this approach also does the same thing, However i havent seen any logic behind this kind of approach, since we might populate assertPage in assertion area, this might look like not clean code, i do believe using text() still would bring the implementation more sufficient, Am not objecting the idea, but am trying to look at the idea that is going to be more appropriate

kdaud: Since we are dealing with loading pages, it would better to rather affirm the page is loaded  successfully using `assertPage() `than using the `assertText()` on the page. If we use texts, then the tests will fail the run in case;
1. the text in the assertion has been changed by UI designers.
2. the locale is changed from English.

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
```diff
@@ -42,13 +42,13 @@ public void loadAllergiesPage() {
...
 	public void systemLoadsAllergiesPage() {
-		assertEquals(getElement(patientHeaderId).getText(),getElement(patientHeaderId).getText());
-		assertTrue(textExists("Allergies"));
+		assertPage(allergyPage.waitForPage());
 	}
 
 	@And("a user clicks No Known Allergy button")
 	public void addNoKnownAllergy() {
 		allergyPage.addNoKnownAllergy();
+		allergyPage.waitForPage();
```
sherrif10, line 51: is this line neccesary here

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
```diff
@@ -59,6 +59,7 @@ public void systemAddsNoKnownAllergies() {
 	@And("a user clicks Remove No Known Allergy icon")
 	public void removeNoKnownAllergy() {
 		allergyPage.removeNoKnownAllergy();
+		allergyPage.waitForPage();
```
sherrif10, line 62: Same here

src/test/java/org/openmrs/contrib/qaframework/automation/PageRejectedException.java
```diff
@@ -27,5 +27,4 @@ public PageRejectedException(String message, Throwable cause,
 			boolean enableSuppression, boolean writableStackTrace) {
 		super(message, cause, enableSuppression, writableStackTrace);
 	}
-
```
sherrif10, line 30: Do we also need empty line here

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
kdaud: adding/removing the allergy some times takes longer and often fails via Bamboo. Invoking waitForPage() gives a room for the operations to be successful before the test performs the next action. Have been observing Bamboo builds failures concerning these operations and we would need all the flaws within these tests fixed so that we the module is stable.

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
kdaud: refer to my comment [here](https://github.com/openmrs/openmrs-contrib-qaframework/pull/256#discussion_r763906928)

src/test/java/org/openmrs/contrib/qaframework/automation/PageRejectedException.java
kdaud: > Do we also need empty line here

we do not need it and its why is removed!

sherrif10 (approved):
Thanks @kdaud

src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
dkayiwa: @kdaud what happens when the page loads without the expected test? Will the test still fail as expected?

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
dkayiwa: @kdaud which action fails when you do not wait? I thought you would wait before the action that fails. Not after the previous one.

dkayiwa:
@kdaud i have done a quick review of this. But am more inclined to @sherrif10 's feelings about these changes. So would need a convincing explanation to think otherwise.

src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
kdaud: > @kdaud what happens when the page loads without the expected test? Will the test still fail as expected?

Have stepped back over the question and realistically the test will pass without the expected test. This would mean resorting to asserting texts on the page however, am skeptical  over asserting tests via texts, in case the locale is changed from English all the tests will fail. Unless we assume this scenario not to happen any way.

src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
kdaud: adding known allergy seems to take longer to perform the action so as the next operation to be effective.

kdaud:
@dkayiwa @sherrif10 your thoughts are worth! The middle ground is to have these tests hard enough not to break so as to serve the intended purpose. Let me know about you thoughts before we may qualify this commit for a revert action.

sherrif10:
@kdaud. To make things more precise, I would assert text instead of a
waiting of a page. These make even the reader to clearly telll what is the
expectations.

src/test/java/org/openmrs/contrib/qaframework/automation/ConditionsSteps.java
dkayiwa: Is asserting text the only way we can reliably locate it?

kdaud: @sherrif10 is there an alternative for asserting texts for loading system page?

kdaud: > Is asserting text the only way we can reliably locate it?

@dkayiwa have not yet known another way but hope will get a better way, for now we can resort to asserting texts.

kdaud:
will revert this commit before Friday ends in case no one objects.
cc: @sherrif10  @dkayiwa

dkayiwa:
No objection from me.

sherrif10:
Me too

kdaud:
Have just noticed that the changes in this commit do not require Revert action since they were made in duplicate files which are being addressed in `pull/265`, so they won't appear in the repo once the pull is merged !!

sherrif10:
> duplicate files which are being addressed in `pull/265`,

Is this already merged or still in progress.

dkayiwa:
> they were made in duplicate files

Here comes another problem of duplicating files. ☺️

kdaud:
> Is this already merged or still in progress

still in progress

kdaud:
> Here comes another problem of duplicating files

Its another C-19 of type. We have other duplicates in other repos that need to address ASAP cc: @sherrif10

Irenyak1:
@kdaud , @sherrif10 ,  @dkayiwa ,  from this [comment](https://github.com/openmrs/openmrs-contrib-qaframework/pull/256#issuecomment-988907267), do we go with asserting text as opposed to pages? Is this our agreed stand?

PR #267, opened 2021-12-09
RATEST-280: Switch Bamboo to use local instance when running tests

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-280)

Description -> Switch Bamboo to use local instance when running tests

### DIFF
--- a/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
+++ b/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
@@ -1,8 +1,9 @@
-webapp.url=https://qa-refapp.openmrs.org/openmrs
+webapp.url=http://localhost:8080/openmrs
 login.username=admin
 login.password=Admin123
 login.location=Pharmacy
 webdriver=chrome
-login.auto=false
+login.auto=true
 headless=true
-webdriver.gecko.driver=firefoxdriver/linux/geckodriver
+db.host=openmrs
+webdriver.chrome.driver=chromedriver/linux/chromedriver


kdaud:
@sherrif10 take a look on the commit for review.

sherrif10:
Any good reason for this change.

qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
```diff
@@ -1,8 +1,9 @@
-webapp.url=https://qa-refapp.openmrs.org/openmrs
+webapp.url=http://localhost:8080/openmrs
```
dkayiwa, line 1: Does this change mean that one needs to have a local instance before they can run the tests successfully?

sherrif10: Lately in september , with intellisoft, we used to use dockerised instance to spin up the instance,however it wasnt reliable especially if a fix is made with in the current module. probably @kdaud . You could help us understand the big reason behind this.

kdaud:
> Any good reason for this change.Is there any failure for using qa server.

@sherrif10 checkout the comments on the [ticket](https://issues.openmrs.org/browse/RATEST-280) to see the reason of switching bamboo to a local instance.

qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
kdaud: > Does this change mean that one needs to have a local instance before they can run the tests successfully?

no !!

Because the `test-bamboo.properties` file is storing settings for tests to use when running on ci(Bamboo). Properties that are governing running the tests locally are residing in _test.properties_ file whose default url points to [qa- instance](https://github.com/openmrs/openmrs-contrib-qaframework/blob/853dfdd339baac31738988d01f466d85724171f5/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test.properties#L1) however, one can change the url to point to a local instance in case has one and opts to use it since it does not depend on the internet connection.

sherrif10:
@kdaud  Have left a quick comment. Feel free to look through.

kdaud:
@sherrif10 checkout the recent comment have added on the ticket.

sherrif10:
1 should work right since it is 2.x ref app? avoid the legacy ui completely except for critical workflows. On contrary critical workflows, i would term this as a critical workflow. This have ever happened on FormTests, manageAccountsTests.

qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
kdaud: > we used to use dockerised instance to spin up the instance,however it wasnt reliable especially if a fix is made with in the current module

With the shared experience on the subject has made me give a thought again if the dockerized instance may not be stable and this is not the direction we would like to take. Not  sure whether our dockerised instance is stable however, its the one we are [using](https://github.com/openmrs/openmrs-contrib-qaframework/blob/853dfdd339baac31738988d01f466d85724171f5/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-local.properties#L1) to run tests in github action and has never disappointed us ever since it was switched in June 2021.

dkayiwa: > `test-bamboo.properties` file is storing settings for tests to use when running on ci(Bamboo)

Does bamboo have an instance running at localhost:8080?

sherrif10:
@kdaud  Sure, i think he is open with the preferred approach. Hope thats the comment and i guess its too direct to use second approach to keep  things instread run on ci

kdaud:
@sherrif10 do you think his response was on the talk post subject?  or on the two suggestions I attached on the ticket?

qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
dkayiwa: Then why did you change the bamboo settings to localhost:8080?

kdaud: > why did you change the bamboo settings to localhost:8080

I had thought localhost:8080 would work since have seen this being te default port when installing a local instance is  `8080` using SDK.

@dkayiwa does this mean the port for `localhost` for Bamboo is a different one?

dkayiwa: Do you have any instance on bamboo running on localhost?

kdaud: @dkayiwa am not sure whether bamboo runs on localhost, and I have looked around but can't find a documentation that talks about it.

As per this conversation seems like bamboo does not run on local host and so we won't switch test-bamboo.properties settings. 

I will then opt for the first alternative I attached on the ticket for creating random values on fields that are causing test failure.

sherrif10:
@kdaud As long as you keep things maintainable and reusable, changing to test.bamboo.properties to spin on localhost isn't the right approach. Again on the work of UserAccount. go where things are achievable, Perhaps we have been doing this testing on legacy ui for some good time so I believe using that part would work for us still.

kdaud:
Will have this PR closed based on the deduction [here](https://github.com/openmrs/openmrs-contrib-qaframework/pull/267#discussion_r766158806) !

PR #283, opened 2021-12-20
RATEST-285: Getting rid of xpath selectors from the page classes

mherman22:
**Ticket ID** : https://issues.openmrs.org/browse/RATEST-285

**Objective** : To get rid of xpath selectors in the page classes and the entire qa codebase and replace them with other Selenium classes provided for in [By](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/By.html).


Ndacyayisenga-droid:
@mherman22 could trigger an empty commit  and rebuild this

sherrif10:
i would have ignored this kind of work because it will bring more errors as possible as it can. Solve issues that might have resulted from your ticket but not rather the entire framework.

mherman22:
> i would have ignored this kind of work because it will bring more errors as possible as it can. Solve issues that might have resulted from your ticket but not rather the entire framework.

That's why I halted it cuz I felt you might not be comfortable. I think we will solve the breakages as they come due to xpaths but as of now (as you said) , lets leave them since they're still working but let it be advised to those who are writing E2E workflow tests to try as much as possible to avoid xpaths in their work. Let us use **className, linkText, ID, partialLinkText, cssSelector** and all others defined in **Selenium's By**


C : @kdaud,  @sherrif10 ,  @Muta-Jonathan @Ndacyayisenga-droid @Irenyak1 @jwnasambu

PR #295, opened 2021-12-27
RATEST-291:RefApp 2.x Form test is failing on github actions

sherrif10:
https://issues.openmrs.org/browse/RATEST-291


kdaud:
@sherrif10 could there be any reason why instantiating a driver and a page within the [hook](https://github.com/openmrs/openmrs-contrib-qaframework/blob/67663b68d6fd5b2770aea1cab57b33bd568e6fc4/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/FormSteps.java#L39-L40)?

sherrif10:
> why a driver and a page was instantiated within the

We are not instantiating drivers but instead pages. The reason is that we want this page to be instantiated not to throw Null pointer Exception.

kdaud (approved):
cc: @dkayiwa

PR #301, opened 2021-12-30
RATEST-293: Add config to trigger Platform install testing on ci

kdaud:


.github/workflows/platform-install-testing.yml
```diff
@@ -0,0 +1,48 @@
+name: Platform Testing Installation
```
sherrif10, line 1: Do we need this workflow for platform testing installations ?

PR #321, opened 2022-01-18
RATEST: Fix flaw in appointment legacy tests

kdaud:


qaframework-legacy-tests/src/test/java/org/openmrs/contrib/qaframework/legacy/AddEditAppointmentBlockTest.java
```diff
@@ -46,8 +46,9 @@ public void addEditAppointmentBlockTest() throws Exception {
         appointmentBlocksPage.selectLocation(locationName);
         appointmentBlocksPage.clickOnDay();
         appointmentBlocksPage.enterStartTime(startTimeFirstAppointment);
-        appointmentBlocksPage.enterService("gyne");
```
sherrif10, line 49: What happens when this is present.

kdaud: "gyne" is an acronym to mean Gynecology as per the test definition. So the search indexing criteria some times is able to map the acronym to the Service type and some times fails. 

I also noticed from the UI that one has to enter Provider first before entering the service type.

sherrif10: According to your explanation above. it looks fairly enough to be approved

sherrif10 (approved):
Thanks @kdaud

kdaud:
Still looking into why ci fails the plan, will update on the subject. cc: @sherrif10

PR #328, opened 2022-01-24
Dockerize All the Workflows To Use Latest Standalone Driver Images

sherrif10:
https://issues.openmrs.org/browse/RATEST-306
The idea behind this changes is that, its not very well stable and practical to keep changing the web drivers each time they are upgraded. 

so if there could be a way to allow drivers gets spinned up locally(using the default browser you are running) or using default browser the ci is running, This is the idea of introducing docker to keep updating selenium images and spin up the workflows smoothly.


sherrif10:
Closing this and openning a new clean commit.

PR #332, opened 2022-01-27
O3-1030: Update E2E 3.x Login test case error message

pasindur99:
## Purpose
The pull request is to fix the 3. x login tests case and this pull request is related to the issue ticket number [O3-1030](https://issues.openmrs.org/browse/O3-1030).

## Approach
Updated the error message in the login to fit with the new UI validation message.

#### Passed test

image


jayasanka-sack (approved):
Thanks @pasindur99

PR #338, opened 2022-01-30
RATEST: Fix the RefApp 2.x tests failing github actions

kdaud:
Creation of test patient should be invoked after the test logged into the system. However, these tests in context are failing with actions but passing with bamboo. cc: @sherrif10

### DIFF
--- a/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
+++ b/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/AllergiesSteps.java
@@ -37,8 +37,8 @@ public class AllergiesSteps extends Steps {
 
 	@Before(RunTest.HOOK.SELENIUM_ALLERGIES)
 	public void visitDashboard() {
-		testPatient = createTestPatient();
 		initiateWithLogin();
+		testPatient = createTestPatient();
 		findPatientPage = homePage.goToFindPatientRecord();
 		findPatientPage.enterPatient(testPatient.identifier);
 		findPatientPage.waitForPageToLoad();
--- a/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/ClinicalVisitSteps.java
+++ b/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/ClinicalVisitSteps.java
@@ -55,8 +55,8 @@ public class ClinicalVisitSteps extends Steps {
 
 	@Before(RunTest.HOOK.SELENIUM_CLINICAL_VISIT)
 	public void visitHomePage() {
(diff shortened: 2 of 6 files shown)


sherrif10:
@kdaud  . Yes have been walking through how a person is generically created using testData. i do think we might not need to use that approach. Let me re run the workflow again.

kdaud:
My thought is around ensuring the function that creates test patent is invoked when the test is logged into the system and I think that would be the better ideal.

Irenyak1:
@kdaud  @sherrif10  why would the tests fail on creating the test patient?

sherrif10:
Let me quickly fix reporting issue.

Irenyak1:
> 

@kdaud however, there's a great number of tests that are creating a patient before logging. Take a scan and see, they are a handful.

kdaud:
> there's a great number of tests that are creating a patient before logging.

@Irenyak1 apart from those subjected to the changes in this PR, do you mind mentioning others that might have been left out and belong to BDD family?

sherrif10:
On question to put in mind, does the test fail locally if we get rid of
testPatient functionality. If not I would consent with this change.

Irenyak1:
> > there's a great number of tests that are creating a patient before logging.
> 
> @Irenyak1 apart from those subjected to the changes in this PR, do you mind mentioning others that might have been left out and belong to BDD family?

AllergiesSteps
ClinicalVisitSteps
ConditionsSteps
FindPatientSteps
InpatientSteps
VitalsAndTriagingSteps

All those ones are creating a patient before logging in.

kdaud:
@Irenyak1 are the tests you have cited out the ones being addressed in this commit?

kdaud:
> does the test fail locally if we get rid of testPatient functionality

The tests run fail since the test object is being consumed within the tests' implementation.

Irenyak1:
> @Irenyak1 are the tests you have cited out the ones being addressed in this commit?

Oh yeah great they are covered. Then there  are no others left out.

sherrif10:
@kdaud  reports should be working now, it was a simple run of the tests.

kdaud:
@sherrif10 am not sure why am not able to reproduce the current build failure locally, are you able to reproduce it on your side?

sherrif10:
This is passing locally on my side. Am not quite sure why its failing still.Let me re build the erver, reports tend to behave like that,view report button becomes disabled when reports gets errored.

kdaud:
@sherrif10 this test may be randomly failing on github action since we can not reproduce it locally. Have just recently run ci and the test is passing so we may need to have this commit enter into master to have the rest of the test github action as we investigate the test further. Let me know what you think to be the way forward?

sherrif10 (approved):
i agree, have also tested it on master branch and its passing. so i agree to merge it and followup on the breakages perhpas

PR #349, opened 2022-02-07
RATEST-293: Add platform upgrade set up wizard tests 

kdaud:


dkayiwa:
What problem are you fixing with this?

kdaud:
This [commit ](https://github.com/openmrs/openmrs-core/commit/b0d8a20291bc60abd1b3180302b8147f26921281) brings in the header page to display the text `Log in as a User that has the System Developer role to continue.` so have added an assertion method that confirms the text is displayed during the upgrade set up wizard.

qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/UpgradeSteps.java
```diff
@@ -29,7 +38,15 @@ public void finish() {
 		complete();
 	}
 
-	@Given("User enters credentials")
+	@Given("User is on login page")
+	public void visitLoginPage() {	
+		assertTrue(textExists(HEADER_TEXT));
+		assertTrue(textExists(SUPER_USER));
+		assertTrue(textExists(OPENMRS_VERSION));
```
dkayiwa, line 45: Will this test pass when the openmrs version is 3.x?

dkayiwa:
Besides the upgrade, is the initial setup also tested with these changes?

qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/UpgradeSteps.java
kdaud: > Will this test pass when the openmrs version is 3.x?

no,

however, one challenge I got in testing out the version is that OpenMRS version evolves over time so we are currently in 2.x version and  when the platform version reaches 3.x then the test will be adjusted accordingly. 

@dkayiwa do you have a suggestion on the subject?

dkayiwa: @kdaud Why do you increase our maintenance burden?

kdaud: > @kdaud Why do you increase our maintenance burden?

@dkayiwa your question is too smart. Well, increasing the maintenance burden is not a good idea.

dkayiwa: I guess you can now come up with a smart solution! 😊

dkayiwa:
So if i reverted any line of that commit, will these tests fail?

kdaud:
> So if i reverted any line of that commit, will these tests fail?

Apart from this [line](https://github.com/openmrs/openmrs-core/blob/96e0ae83fc5c6da3521a99a27c43ddfb720b2e27/web/src/main/java/org/openmrs/web/filter/update/UpdateFilterModel.java#L37) which I find having the no effects on the UI, the rest of the lines when reverted the test fails. With the line mentioned above, it is supposed to show the footer on the page which is not true.

qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/InitialSetupSteps.java
```diff
@@ -15,6 +15,9 @@
 import org.junit.Assert;
 import org.openmrs.contrib.qaframework.helper.InitialSetupPage;
 import org.openmrs.contrib.qaframework.helper.TestBase;
+import org.openqa.selenium.By;
+import org.openqa.selenium.NoSuchElementException;
```
dkayiwa, line 19: Is this exception in use?

kdaud: > Is this exception in use?

Have done the needful in the next subsequent commit with id `9779260`

dkayiwa:
@kdaud i do not see the platform tests included in the above checks.

kdaud:
> @kdaud i do not see the platform tests included in the above checks.

@dkayiwa all the platform tests were configured not to be triggered by any PR. However, they all run on every merge made on master. So that is the reason why they are not shown in the checks. What I often do for testing out the platform tests is to run a plan for each test against the branch via my fork.

dkayiwa:
Are you doing any more work on this?

kdaud:
If the ideal is changing their original set up(for not being triggered on the PR ) then that will call for additional config in their respective workflows to be triggered on every PR made on master.

kdaud:
> Are you doing any more work on this?

We are good to go unless we are to change the original set up for platform tests ie making the tests be triggered on every PR!

dkayiwa:
@kdaud do the platform tests run immediately on merging?

kdaud:
> @kdaud do the platform tests run immediately on merging?

yes they do. take a look at the current master plan
Screenshot from 2022-02-15 16-18-38

dkayiwa:
How about when i make changes to openmrs-core? Do they run immediately on commit?

kdaud:
> How about when i make changes to openmrs-core? Do they run immediately on commit?

This is something am sure of, because what I know for now is that run every merge made on qaframework module.

dkayiwa:
> am sure of

Is this a typo?

kdaud:
> > am sure of
> 
> Is this a typo?



It is! I meant `am not sure of`


Am looking into whether they are triggered or not on a change made on the core, and will respond to you ASAP.

dkayiwa:
I remember as if some one said they run once a day. I feel that this should be changed to every commit of openmrs-core

kdaud:
> I remember as if some one said they run once a day.

I also think the person was pointing to the Bamboo plan that at-least run once every day. Usually this configured ci plan happens at 4:00 AM EAT.

> I feel that this should be changed to every commit of openmrs-core

This is a good idea, my understanding  on the subject is to be achieve the goal vai Bamboo config,

dkayiwa:
How about the ones on this dashboard? https://github.com/openmrs/openmrs-contrib-qaframework

Will they immediately fail if i revert that commit in openmrs-core?

kdaud:
> How about the ones on this dashboard? https://github.com/openmrs/openmrs-contrib-qaframework
> 
> Will they immediately fail if i revert that commit in openmrs-core?

AFAICT those on the dashboard won't fail until a merge is done on the qaframework repo to trigger them to run. This is when the tests will detect a change on the openmrs-core and this means a change done on the core is detected by the tests later.

dkayiwa:
Can't you make them run immediately on openmrs-core commit?

kdaud:
> Can't you make them run immediately on openmrs-core commit?

Have started off with the task!

PR #352, opened 2022-02-08
Switch tests to run on the demo server in github actions

kdaud:


qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-local.properties
```diff
@@ -1,9 +1,9 @@
-webapp.url=http://localhost:8080/openmrs
+webapp.url=https://demo.openmrs.org/openmrs
```
sherrif10, line 1: Hello @kdaud  any reason for this change. we had agreed to use local host

kdaud: I was testing carrying out a demo on a certain use case. So the commit has no harm on master!

PR #355, opened 2022-02-09
O3-1057 Move User Settings E2E test to the dockerized environment

99sumedha:
The purpose of this PR is to fix [O3-1057](https://issues.openmrs.org/browse/O3-1057).

Approach
Run 3.x user settings E2E test in a dockerized environment.
Removed location Pharmacy and Replace Inpatient Ward

### DIFF
--- a/.github/workflows/refapp-3x-settings.yml
+++ b/.github/workflows/refapp-3x-settings.yml
@@ -1,9 +1,9 @@
 name: RefApp 3.x User Settings
 on:
   push:
-    branches: [ master ]
+    branches: [ master, dockerize-3.x ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, dockerize-3.x ]
   repository_dispatch:
     types: [ qa ]
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


99sumedha:
Done @jayasanka-sack

99sumedha:
@jayasanka-sack It fails cuz of a bug in the system. After those two are fixed the test will pass

99sumedha:
https://talk.openmrs.org/t/fixing-failing-3-x-e2e-tests-in-qa-dashboard/35487/17?u=sumedha @jayasanka-sack

.github/workflows/refapp-3x-settings.yml
```diff
@@ -14,7 +14,14 @@ jobs:
       run:
         working-directory: ./qaframework-bdd-tests
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout qaframework
+        uses: actions/checkout@master
+        with:
+          repository: ${{github.repository}}
```
pasindur99, line 20: There's no need to add this section to the workflow because `repository: ${{github.repository}}` contains a default value.

pasindur99 (changes requested):

.github/workflows/refapp-3x-settings.yml
99sumedha: Sure, should I remove it?

jayasanka-sack:
The setup isn't viable to address challenges. We should work on a different approach. Therefore, closing the PR.

PR #359, opened 2022-02-10
Bump selenium version from 4.1.2 to 4.0-alpha-6

kdaud:

PR #362, opened 2022-02-14
Patch the selenium version and geck driver version

sherrif10:
Hey @dkayiwa  @kdaud  i would think us to revert back selenium version and gecko driver such that we can settle things . i remember reports workflow is the one that resulted into such failures. Let me know what you think.


kdaud:
@sherrif10 following the conversation at [7f1f3b5](https://github.com/openmrs/openmrs-contrib-qaframework/pull/359) we agreed not to downgrade selenium due to the reasons given there. are you suggesting otherwise?

sherrif10:
Thanks @kdaud . Yes am bit convinced that we should downgrade the selenium version back to its current version first . then handle those errors per error.

kdaud:
did you take a close look at the conversation on the commit shared above?

sherrif10:
Thats cool. Lets wait for @dkayiwa  response on that matter.

PR #363, opened 2022-02-14
Switch tests on git-hub actions to run local host instead of qa instance

kdaud:


kdaud:
@sherrif10 Report test fails on local host and we would prefer to switch from qa instance to fix platform tests, do you any recommendation?

dkayiwa:
@kdaud can i look at the error log when pointing to the qa instance?

sherrif10:
The problem is, when a system had finished covering the reports, we are supposed to view reports, so the driver dont click on view report link on local lost  but does it on qa instance.

dkayiwa:
> The problem is, when a system had finished covering the reports, we are supposed to view reports, so the driver dont click on view report link on local lost but does it on qa instance.

Can i look at the full log?

kdaud:
> @kdaud can i look at the error log when pointing to the qa instance?

The RefApp 2.x Report test passes well on the qa instance but [fails](https://github.com/openmrs/openmrs-contrib-qaframework/runs/5182086429?check_suite_focus=true#step:7:12942) on the local host at the last [step](https://github.com/openmrs/openmrs-contrib-qaframework/blob/35ed46e4ba3c55bb4ceed0cfcb7576022aa352dd/qaframework-bdd-tests/src/test/resources/features/refapp-2.x/015-reports/reports.feature#L17) implemented [here ](https://github.com/openmrs/openmrs-contrib-qaframework/blob/35ed46e4ba3c55bb4ceed0cfcb7576022aa352dd/qaframework-bdd-tests/src/test/java/org/openmrs/contrib/qaframework/automation/ReportSteps.java#L85-L88). The platform tests passes well on the local host and fails on the qa instance with these [logs](https://github.com/openmrs/openmrs-contrib-qaframework/runs/5182829592?check_suite_focus=true)

kdaud:
Currently github actions is running tests on qa instance however, all the platform tests fail the plan. For these tests to pass, they should run on the local instance which is suggested in this commit.

dkayiwa:
@kdaud my understand is that the platform tests are about the openmrs installation and upgrade processes which is not what the qa instance is running.

kdaud:
> @kdaud my understanding is that the platform tests are about the openmrs installation and upgrade processes which is not what the qa instance is running. So they cannot be together with the rest of the tests. Do i make sense?

@dkayiwa your understanding is right!

Now that platform tests are residing in the same sub module with bdd tests, this makes it hard to separate their respective instances for which each can run against. If we are to have them with distinctive instances, then the work around would be refactoring platform tests to run in a different sub-module. Is this a better move?

sherrif10:
@dkayiwa  here is the log of failing report. https://pastebin.com/fDwWsiYX  .

dkayiwa:
@kdaud what changed? Were they originally set up this way? For they were running fine.

kdaud:
> @kdaud what changed? Were they originally set up this way? For they were running fine.

Originally all the tests(bdd, legacy, platform) were running well on local host on github actions until RefApp 2.x Report tests(bdd test) started to fail and this [commit](https://github.com/openmrs/openmrs-contrib-qaframework/commit/afbed520d89552b1f2d1cfe5a4936ecc88f74b47) changed the server to qa since Report test was passing with it. However, this caused all the platform tests to fail since qa is not running  what platform installation and upgrade process is all about.

sherrif10:
According to the discussions we had on slack, am sure @kdaud  you have been seeing them. Let's merge this then fix report workflow separately

sherrif10:
@kdaud  do you mind first disabling the last scenario of reports which is failing to get triggered first.

qaframework-legacy-tests/src/test/resources/org/openmrs/uitestframework/test-local.properties
```diff
@@ -7,4 +7,4 @@ login.auto=true
 headless=true
 db.host=openmrs
 webdriver.gecko.driver=firefoxdriver/linux/geckodriver
-includes.csrftoken=true
+includes.csrftoken=false
```
sherrif10, line 10: this should still be true not false

kdaud:
> According to the discussions we had on slack, am sure @kdaud you have been seeing them

Have added a comment on slack after following up the conversation.

sherrif10:
Hope the conversation was so fruitfull.

kdaud:
> @kdaud do you mind first disabling the last scenario of reports which is failing to get triggered first.

@sherrif10 are you making a follow up on this? IMO: It would be better to fix what is failing than just disabling stuff.

sherrif10:
> It would be better to fix what is failing than just disabling stuff.

Do you mind doing the needful in this Pr and save time.

kdaud:
> you would wish me to fix it separately.

Feel free to fix. The commit here is switching actions back to its original setting and its not a good practice to mix up commits that address different things.

qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-local.properties
```diff
@@ -1,10 +1,10 @@
-webapp.url=https://qa-refapp.openmrs.org/openmrs
+webapp.url=http://localhost:8080/openmrs
 login.username=admin
 login.password=Admin123
 login.location=Pharmacy
 webdriver=firefox
 login.auto=false
 headless=true
-#db.host=openmrs
+db.host=openmrs
```
sherrif10, line 8: Is this also needed as part of the commit

kdaud: the db server need to be set when running on local host.

kdaud:
> i was suggesting to comment out the last report.feature which is not getting triggered

Pushing work a head is not a good deal unless its inevitable. @sherrif10 are you going to follow up on this and perhaps create a ticket to fix the issue?

sherrif10:
> > i was suggesting to comment out the last report.feature which is not getting triggered
> 
> Pushing work a head is not a good deal unless it's inevitable. @sherrif10 are you going to follow up on this and perhaps create a ticket to fix the issue?

All is well @kdaud . will fix it immediately.

sherrif10 (approved):
Thanks @kdaud

sherrif10:
Merging this the awaiting one.

kdaud:
> > > i was suggesting to comment out the last report.feature which is not getting triggered
> > 
> > 
> > Pushing work a head is not a good deal unless it's inevitable. @sherrif10 are you going to follow up on this and perhaps create a ticket to fix the issue?
> 
> All is well @kdaud . will fix it immediately.

@sherrif10 how far with the fix?

kdaud:
> > > i was suggesting to comment out the last report.feature which is not getting triggered
> > 
> > 
> > Pushing work a head is not a good deal unless it's inevitable. @sherrif10 are you going to follow up on this and perhaps create a ticket to fix the issue?
> 
> All is well @kdaud . will fix it immediately.

@sherrif10 how far with the fix?

sherrif10:
which fix is that, i dont remember it.

kdaud:
> @kdaud personally it would be much better to see this commit pass all the workflows including reports, so thats why initially i was suggesting to comment out the last report.feature which is not getting triggered.

Does the comment above answers your question?

PR #379, opened 2022-02-24
Fix the ocl subscription test

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-313)

### DIFF
--- a/.github/workflows/ocl-subscription-module.yml
+++ b/.github/workflows/ocl-subscription-module.yml
@@ -15,6 +15,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./qaframework-bdd-tests
     steps:
       - uses: actions/checkout@v2
       - name: Using Node.js
@@ -29,8 +32,6 @@ jobs:
           docker-compose -f docker-compose-refqa.yml up -d
       - name: wait for openmrs instance to start
         run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 1; done
-      - name: Navigate into qaframework-bdd-tests
-        run: cd qaframework-bdd-tests
       - name: Run ocl subscription module
         env:
           CYPRESS_OCL_USER_NAME: ${{secrets.OCL_USER_NAME}}
@@ -44,4 +45,4 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Screen recordings of failed tests
-          path: cypress/videos/oclsubscriptionmodule
+          path: qaframework-bdd-tests/cypress/videos/oclsubscriptionmodule
--- a/qaframework-bdd-tests/src/test/resources/features/oclsubscriptionmodule/ocl-subscription-module.feature
+++ b/qaframework-bdd-tests/src/test/resources/features/oclsubscriptionmodule/ocl-subscription-module.feature
@@ -1,0 +1,20 @@
+Feature: Import Dictionary via subscription module
+
+  Scenario: The user should copy the subscription url from the Dictionary manager
+    Given the user goes to the Dictionary Manager App
+    When the user logs into the Dictionary Manager
+    And the user goes to the dictionary details page
+    When the user clicks the more actions button
+    And the user selects the "Copy Subscription URL" menu list item
+    Then the user should copy the subscription url
+
+  Scenario: The User should be able to install the dictionary in openmrs
+    Given the user goes to the openmrs application
+    When the user logs into the openmrs Ref App
+    And the user visit the OCL module page
+    When the user pastes the subscription url and user API token
+    And the user saves changes
+    Then the user should see the import button
+    When the user clicks the import button
+    Then the dictionary should be imported
+    And the data should be correct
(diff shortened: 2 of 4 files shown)


qaframework-bdd-tests/src/test/resources/features/oclsubscriptionmodule/ocl-subscription-module.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Import Dictionary via subscription module
+
+  Scenario: The user should copy the subscription url from the Dictionary manager
+    Given the user goes to the Dictionary Manager App
```
sherrif10, line 4: Do we have step definition class for these guys.

sherrif10:
Hey i can still see some failures probably you might need to check with ocl server.

qaframework-bdd-tests/src/test/resources/features/oclsubscriptionmodule/ocl-subscription-module.feature
kdaud: https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js

.github/workflows/ocl-subscription-module.yml
```diff
@@ -39,4 +40,4 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Screen recordings of failed tests
-          path: cypress/videos/oclsubscriptionmodule
+          path: qaframework-bdd-tests/cypress/videos/oclsubscriptionmodule
```
ibacher, line 43: The video that gets uploaded here is pretty much useless. Would it also be possible to attach the screenshots of the failed tests?

ibacher:
At least the failures have moved onto something else... I'm not entirely sure that `export const VARIABLE = Cypress.env("VARIABLE");` is the right pattern to use to get these values. It might be better to either wrap it as a command or at least a function, e.g.,

```
export const getOclUserName = () => Cypress.env("OCL_USER_NAME");
```

Alternatively:

```
Cypress.Commands.add("oclUserName", () => Cypress.env("OCL_USER_NAME"));
```

Which is a big more convenient to use:

```
cy.get("#username").type(cy.oclUserName());
```

qaframework-bdd-tests/src/test/resources/features/oclsubscriptionmodule/ocl-subscription-module.feature
```diff
@@ -0,0 +1,20 @@
...
+    When the user logs into the Dictionary Manager
+    And the user goes to the dictionary details page
```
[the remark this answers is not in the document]
kdaud: The second scenario does not use these steps because it uses a different server(RefApp server).

.github/workflows/ocl-subscription-module.yml
kdaud: > Would it also be possible to attach the screenshots of the failed tests?

The build plan summery provided by git-hub actions is good enough to show which tests have failed and those that have passed.https://github.com/openmrs/openmrs-contrib-qaframework/runs/5355693711?check_suite_focus=true#step:7:44

ibacher: Yes, it shows which tests failed. What it doesn't show is *why* those tests failed. I can't tell which step in the flow causes the error: "CypressError: `cy.type()` cannot accept an empty string. You need to actually type something." nor can I see which field it's trying to type in. A screenshot would solve both of those problems (the video doesn't actually capture anything useful).

sherrif10:
@kdaud  Does this work locally and fail on Github actions.

kdaud:
> Does this work locally and fail on Github actions

This test has the same issue locally and most probably the issue may be coming from using a secrete username for ocl server.

.github/workflows/ocl-subscription-module.yml
kdaud: Import Dictionary via subscription module -- The User should be able to install the dictionary in openmrs (failed)

@ibacher the credentials `OCL_USER_NAME`, `OCL_USER_PASSWORD,` `OPENMRS_USER_NAME`, `OPENMRS_USER_PASSWORD` and `OCL_USER_TOKEN `are failing to be imported into the cypress steps definition hence CypressError: `cy.type()`

Have instead used the credentials directly in the steps file( though its not good to expose these credentials in steps definition). I know these credentials were set as secret values however, could there be a possibility that these secrete credentials are expired and need to be updated?

kdaud:
Not sure why `Cypress.env("VARIABLE")` fails to import the variable hence throwing the logs: `cy.type() can not accept an empty string`.
jj

kdaud:
The tests are passing when variables are passed directly as strings something like;
` cy.get("#username").type("username");`

sherrif10:
@kdaud  is this still needed.

kdaud:
The OCL Subscription test is failing to import the credentials for open concept and refapp server throwing Cypress error:` cy.type() can not accept an empty string.` as shared in the above.

One approach would be passing these variables as strings, have made `pull/399` with the suggested solution and the test is passing Actions.

kdaud:
Fixed at https://github.com/openmrs/openmrs-contrib-qaframework/commit/1d8c86065d2d1fc33df7648b5cd5cfa520f13220

PR #384, opened 2022-03-08
Segregate which workflows get run based on which files are touched

ibacher:
What this does is determine (for PRs only) which jobs should be run based on which files were touched. I noticed that, e.g., when looking at [this PR](https://github.com/openmrs/openmrs-contrib-qaframework/pull/379) none of the 3.x jobs were being run... never mind the actual job that was supposed to fix. We should actually probably limit this down further for pushes etc. as we can only have 20 jobs run at one time, so, e.g., on push other jobs aren't necessarily being run.

PS This PR itself should have no jobs run since it didn't touch any of the files.

### DIFF
--- a/.github/workflows/ocl-subscription-module.yml
+++ b/.github/workflows/ocl-subscription-module.yml
@@ -4,9 +4,14 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
+    paths:
+      - "qaframework-bdd-tests/cypress/**/*.*"
+      - "qaframework-bdd-tests/cypress.json"
+      - "qaframework-bdd-tests/package.json"
   repository_dispatch:
     types: [qa]
   workflow_dispatch:
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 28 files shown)


kdaud:
Have skimmed through the commit and makes sense altogether. One thing would be adding the path for package.json in all the RefApp 2.x tests .yml files. Something like;

```js
paths:
      - "qaframework-bdd-tests/src/**/*.*"
      -  "qaframework-bdd-tests/package.json"
```

sherrif10 (commented):
Thanks @ibacher  . i definately agree with this approach because it saves resources and improves maintanability of the product.

sherrif10:
cc @ibacher  Looking forward to hear from you.

kdaud:
@sherrif10 we need these changes in our master so I suggest we get this PR merged then we can make follow  commit if needed. @ibacher seems to be busy and may take a while to come on board so we can test out these changes and make a follow up commit if required. 

Let me know what you think?

ibacher:
@kdaud I've included you're suggestion in the 2.x packages in the actual repo commit.

kdaud:
Thanks @ibacher . I also made a [comment](https://github.com/openmrs/openmrs-contrib-qaframework/pull/379#issuecomment-1065115558) in regard to fixing ocl subscription test.

PR #392, opened 2022-03-15
Switch RefApp 3.x tests to run on o3.openmrs.org server

kdaud:


dkayiwa:
Is this supposed to run the RefApp 2.x tests?

Ndacyayisenga-droid:
@kdaud  did you try to investigate why this is failing?

jayasanka-sack:
Hi @kdaud,
Thanks for the PR.

Currently, these tests are failing due to many reasons. 
ex:
- UUIDs change when changing the server (there are some hardcoded UUIDs)
- The UI takes a significant amount of time to load (because of the offline mode). Sometimes even it fails to load
- It is running on a deployed server. Hence, data mutations affect existing tests.

We might need structural changes to address those problems. Hoping we can come up with proper solutions with the 3.x E2E GSoC project.

kdaud:
> We might need structural changes to address those problems. Hoping we can come up with proper solutions with the 3.x E2E GSoC project.

Sounds a good plan!

I disabled all the 3.x tests since they were failing on every push done on master, so when work around them kicks off feel free to enable them to run on plans. cc: @jayasanka-sack

PR #399, opened 2022-03-21
Resurrect ocl subscription test

kdaud:


sherrif10:
Lets create  ticket for this issue.

kdaud:
Check my recent comment on `pull/379`

kdaud:
@sherrif10 when we pass the server's credentials directly as strings the OCL Subscription Module test passes actions.

sherrif10:
> Check my recent comment on `pull/379`

can you add the link to it here.

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
```diff
@@ -60,13 +53,13 @@ Then('the user should copy the subscription url', () => {
 });
 
 When('the user visit the OCL module page', () => {
-  cy.visit('http://localhost:8080/openmrs/owa/openconceptlab/index.html#/subscription')
+  cy.visit('https://demo.openmrs.org/openmrs/owa/openconceptlab/index.html#/subscription')
```
sherrif10, line 56: Any reason why you are using demo instead of local host.

kdaud:
> can you add the link to it here

https://github.com/openmrs/openmrs-contrib-qaframework/pull/379

sherrif10:
@kdaud  isn't this this test supposed to be segregated. I hope to see only one workflow getting build

sherrif10:
The recent commit @ibacher  shared was to help us run workflows per a given workflow. We don't need to see all workflows getting triggered. So I still don't get it when I see many workflows getting triggered.

kdaud:
The reason why it triggers tests for 2.x is that have added a feature file within the path "qaframework-bdd-tests/src/**/*.*" hence triggering the 2,x tests to run. If the feature file was not added then your expectation would be true. Just in case you want to verify I can temporarily delete out the feature file added in that path and then only OCL Subscription Module test will run.

kdaud:
@sherrif10 what do you think of this PR?

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
```diff
@@ -25,15 +18,15 @@ Given('the user goes to the openmrs application', () => {
...
+  cy.get("#username").type("openmrs");
+  cy.get("#password").type("Openmrs123");
   cy.get("#login-form form").submit();
   cy.url().should('contain', 'https://openmrs.staging.openconceptlab.org/user/collections/');
 });
 
 When('the user logs into the openmrs Ref App', () => {
-  cy.get("#username").type(OPENMRS_USER_NAME);
-  cy.get("#password").type(OPENMRS_USER_PASSWORD);
+  cy.get("#username").type("admin");
```
sherrif10, line 28: @kdaud   why this change. what happens when its left the way it was initially.

sherrif10:
@kdaud  i can see some un addressed comments.

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
kdaud: The server credentials fails to be exported hence throwing Cypress error: `cy.type() can not accept an empty string`, on the the other hand as suggested in the PR when the credentials are passed directly as strings the test passes.

kdaud:
> @kdaud i can see some un addressed comments.

Do you mind pointing me to the un addressed comment?

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
sherrif10: Any response on the above comment.

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
sherrif10: > `cy.type() can not accept an empty string`,

You mean even when you  equate that string  to a value, something like String` USER_NAME  = "Admin"`, it still fails ?

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
kdaud: Sure I reverted the server back to the local instance that is why it's showing `Outdated` label.

qaframework-bdd-tests/cypress/integration/cucumber/step_definitions/oclsubscriptionmodule/ocl-subscription-module.js
kdaud: The credentials are set in command.js file
```js
export const OCL_USER_NAME = Cypress.env('OCL_USER_NAME');
export const OCL_USER_PASSWORD = Cypress.env('OCL_USER_PASSWORD');
export const OPENMRS_USER_NAME = Cypress.env('OPENMRS_USER_NAME');
export const OPENMRS_USER_PASSWORD = Cypress.env('OPENMRS_USER_PASSWORD');
export const OCL_USER_TOKEN = Cypress.env('OCL_USER_TOKEN'
```
which are then imported as 

```js
import { 
  OCL_USER_NAME, 
  OCL_USER_PASSWORD, 
  OCL_USER_TOKEN,
  OPENMRS_USER_NAME, 
  OPENMRS_USER_PASSWORD
} from '../../../../support/commands';
```
Then passed in their respective fields. However they are all failing to be imported hence throwing the error I shared above.

sherrif10: Are they exported, Usually react js, js classes need to be exported first for them to be accessed.

sherrif10: @kdaud  Seems even the export function does not work as expected. Look here https://github.com/openmrs/openmrs-contrib-qaframework/blob/master/qaframework-bdd-tests/cypress/support/commands.js#L10

kdaud: Do you have a suggestion that can use rather?

sherrif10: If this really failed with that initial work, then we can go with direct placing of those attribute.

kdaud: Have not gotten a way of importing them after trying several alternatives I know. If there is one you know then we can use that rather otherwise we can have them passed directly as strings and if we get a way out  then we can do refactor the test.

sherrif10: > we can have them passed directly as strings and if we get a way out then we can do refactor the test.

I agree because i have gone through it microscopically and am conviced this looks good.

sherrif10 (approved):
Thanks @kdaud

sherrif10:
Thanks @kdaud  Hopefuly you will be winking on the build of the OCL part.

kdaud:
@sherrif10  I see Green Lights after the build plan. The test is back to life!
Screenshot from 2022-03-28 12-09-46

sherrif10:
ohh yeah man. Awesome.

PR #423, opened 2022-06-13
RATEST-330: Fix delete resource within rest client call

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-330)

Description -> Fix delete patient within rest client call

### DIFF
--- a/qaframework-bdd-tests/src/main/java/org/openmrs/contrib/qaframework/helper/TestBase.java
+++ b/qaframework-bdd-tests/src/main/java/org/openmrs/contrib/qaframework/helper/TestBase.java
@@ -303,11 +303,13 @@ public class TestBase {
 	 * Delete the given patient from the various tables that contain portions of
 	 * a patient's info.
 	 *
-	 * @param patientInfo containing hhe uuid of the patient to delete.
+	 * @param patientInfo containing the uuid of the patient to delete.
 	 */
 	public void deletePatient(PatientInfo patientInfo) throws NotFoundException {
 		if (patientInfo != null) {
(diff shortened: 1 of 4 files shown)


sherrif10:
@kdaud  Weldone, some errors on All firefox to look at.

kdaud:
@sherrif10 am done with fixing the root cause. So we are good to go -->

dkayiwa:
@kdaud how does just catching and ignoring the exception fix the problem?

kdaud:
Here is the context:

The tests are failing to delete the resource(testPateint) at the end of the run probably due to Rest client call responsible to for the operation. 

> java.lang.IllegalStateException: Delete request failed: https://qa-refapp.openmrs.org/openmrs/ws/rest/v1/patient/b831a4d2-a64f-4c5b-a291-1b85f39272f3

The logs are similar and signals that RestClient.delete() is being invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation. So adding the try catch makes the operation succeed.

dkayiwa:
Shouldn't we just fix the cause of the exception? What you have done in this pull request is taking the whistle blower to prison instead of using their information to find the culprit.

kdaud:
> What you have done in this pull request is taking the whistle blower to prison instead of using their information to find the culprit.

Have embarked on using their information to find the exact culprit and then deal with him as the law states.

sherrif10:
@kdaud Are we still interested in these changes.

kdaud:
The snag is fixed.

PR #436, opened 2022-07-29
Bump selenium version to 4.1.4

kdaud:
Upgrade selenium version to 4.1.4 to attempt fixing the current Bamboo failure.

### DIFF
--- a/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
+++ b/qaframework-bdd-tests/src/test/resources/org/openmrs/uitestframework/test-bamboo.properties
@@ -5,5 +5,5 @@ login.location=Pharmacy
 webdriver=chrome
 login.auto=true
 headless=true
-webdriver.gecko.driver=firefoxdriver/linux/geckodriver
+webdriver.chrome.driver=chromedriver/linux/chromedriver
 includes.csrftoken=true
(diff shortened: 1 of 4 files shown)


sherrif10:
Does it break the ci, is it helping us to resolve bamboo

kdaud:
The failures at Bamboo are too generic that can not give the right cause why all the tests are failing. I think upgrading the selenium version to the latest will give a better Bamboo logs that could be helpful.

Besides, library upgrade are so good to have in our module and am wondering if the we could have openmrs-bot configured to have it help us in upgrading the module dependencies.

qaframework-bdd-tests/pom.xml
```diff
@@ -47,7 +47,7 @@
 		<dependency>
 			<groupId>org.seleniumhq.selenium</groupId>
 			<artifactId>selenium-java</artifactId>
-			<version>4.3.0</version>
+			<version>4.1.3</version>
```
sherrif10, line 50: Any Reason of reverting it back ?

kdaud: Am just testing out something but will get it back to the latest version.

kdaud: Any version higher than 4.1.4 breaks all the tests so have instead upgraded to 4.1.4

PR #450, opened 2022-08-18
Upgrade chrome-driver to v99.0.4844 and set driver-path

kdaud:


sherrif10:
Thanks @kdaud  i can see the ci is happy now, the problem was with chrome driver version.

dkayiwa:
For a build that has taken less than a minute to complete, we are not there yet. He must have just messed with the bamboo config 😊

kdaud:
I had disabled the task that runs the tests and that is why green came up. I notice that when the task is put on the time it takes to build is over 3 hours which is quite abnormal and turns out red. The delay is not caused by any of the tests taking long to build but rather much time is wasted by each of the test when trying to instantiate the chrome driver. None of the tests succeed in instantiating the driver and all the tests fail.

Earlier on I switched Bamboo to use firefox driver and the behavior was the same during the plan.

Tests are running well on github actions on either chrome or firefox but the issues is with Bamboo.

Thought: Most likely the issue is coming from our driver configuration that need some adjustments.

dkayiwa:
@kdaud id you say that the issue is with bamboo, did you confirm that the tests run locally?

kdaud:
Am able to run the tests locally. And as well confirmed the chrome driver on github action using qa-refapp instance via this [commit](https://github.com/kdaud/openmrs-contrib-qaframework/pull/27)

dkayiwa:
How long do they take to run locally? And are you using the chrome or firefox driver?

kdaud:
About the 1 hour : 30 minutes on chrome. Have not tested on firefox.

dkayiwa:
That still sounds too much!

kdaud:
Am using qa-reapp server to run the tests and probably internet speed is probably affecting the performance. I used my  local instance and the time was 40-45 minutes but some tests were failing due to some modules in the server  that are not updated with fixes that were made to the remote repos.

PR #453, opened 2022-09-07
Remove login.auto property that is prohibiting tests login

kdaud:


kdaud:
The `login.auto` property is the root cause of the tests failure in the continuous integration. We can either set it to false which is equivalent to getting rid of the property. This commit will make Bamboo return green builds. cc: @dkayiwa

dkayiwa:
@kdaud where have you been hiding? I thought you had given up on the fellowship.

As for your attempt to fix bamboo, feel free to merge and see.

dkayiwa:
> Haaaa me to , I have asking myself same question, may be the man have been not well.

When one is not well, they communicate to the mentor. Because this did not happen, i thought it was an issue of giving up on the fellowship.

kdaud:
> @kdaud where have you been hiding?

I was attacked by covid-19 and wasn't an easy experience on my side. God reserved my life and recovered from the pandemic!  

> When one is not well, they communicate to the mentor

I remember informing you about what I was going through.

